PR #1, opened 2016-07-28
Added Italian translation

insomniacslk:


app/src/main/res/values-it/strings.xml
```diff
@@ -0,0 +1,95 @@
...
+	<string name="chooser_past">Eventi precedenti:</string>
+	<string name="add_dialog">Aggiungi...</string>
+	<string name="qr_scan">Leggi QR</string>
+	<string name="enter_url">Inserisci URL del programma</string>
+
+	<!-- Reminder -->
+	<string name="soon_in">Prossimamente in</string>
+
+	<!--  Widget -->
+	<string name="widg_longdate">EEEE d MMMM \'at\' HH:mm</string>
```
Wilm0r: Not that I know Italian but I would guess that this 'at' needs to be 'alle' like below?

insomniacslk: Yes, my fault. I did it at the line below but missed it here. PR update is coming :)

PR #2, opened 2016-07-29
Added MOCA 2016 event schedule

insomniacslk:


app/src/main/res/raw/menu.json
```diff
@@ -155,6 +155,40 @@
 					}
 				]
 			}
+		},
+		{
+			"version": 2016072901,
+			"url": "https://moca.olografix.org/moca2016.ics",
+			"title": "MOCA 2016",
+			"start": "2016-07-19",
+			"end": "2016-07-21",
```
Wilm0r: Is that the right date?

(It's not a big deal since Giggity will update these in its database based on your .ics on first load, but still it'd look odd having your event listed as already in the past.)

app/src/main/res/raw/menu.json
```diff
@@ -155,6 +155,40 @@
...
+						"url": "https://moca.olografix.org/it/speaker/",
+						"title": "Speakers"
+					},
+					{
+						"url": "https://moca.olografix.org/it/logistica/",
+						"title": "Logistics"
+					}
+				],
+				"icon": "https://moca.olografix.org/moca2016.png",
+				"rooms": [
```
Wilm0r: Cool, you'll be the first to use that feature. :-D

Wilm0r:
Neat, did you manage to test this by loading the whole thing into Android Studio or so? That's a lot of work, and I should add the ability to load json files with all this metadata from the main menu.

app/src/main/res/raw/menu.json
insomniacslk: ..no, it's not. I've just corrected it, thanks for spotting!

app/src/main/res/raw/menu.json
insomniacslk: I haven't used this feature and I already love it :) Thanks for the advice!

insomniacslk:
No, I haven't tried to load it on Android Studio so it's basically untested. I've just installed it on another laptop and I'm about to try

Wilm0r:
Oh, don't worry, I can do the testing for you, installing Android Studio just to test this is rather overkill.

insomniacslk:
Thanks Wilmer. I'm trying anyway to speed up future testing on my side. It seems to be unhappy with the SDK API 21 (defined in giggity/app/build.gradle) on a recent Android Studio installation, I'll try bumping to the latest ASAP.

Wilm0r:
That's odd, I don't have that problem and I'm pretty sure I have the latest version as well.

Still, that stuff's mostly opaque to me and I just click my way to stuff working so I can go back to doing what I want (write code and run it). Hope I didn't make too big a mess. ;-)

insomniacslk:
It's probably something missing in my installation, I'll sort it out tonight (or if you manage to test it before I do, it's obviously welcome :) )

Wilm0r:
Tested this locally. Although MOCA shows up in the main menu, it ends up showing twice and the metadata (room info and links) don't work, likely because of a mixup of unique IDs/primary key.

My first thought is to blame the .ics bit which IIRC I used to fix by adding the correct id to the seed file, but last time in 2012 your entry was pretty simple as well (b3b3630979be6e8bdc9ca1d09760baf1c49f9452).

I'll look more closely tomorrow maybe.

insomniacslk:
Not too sure about this either. I will try again today

Wilm0r:
Found it. The following line fixes it:

                        "id": "MOCA 2016", // Warning: Specifying ID is not normally necessary!

That id, not the schedule URL, is used as the unique identifier. I should probably remove that behaviour though, since in fact with the most commonly used Pentabarf parser, I'm falling back to a checksum of the URL already since Pentabarf doesn't really have any unique identifier field. Which is probably fine since these things don't really move around that much.

Also, the name of your "ELF Arena" looks quite different in the iCal file, can you fix that to match the regex in the json metadata?

insomniacslk:
Thanks! I finally managed to build the app and could reproduce the problem, but you were faster at nailing it :)

I'm now looking at fixing what you pointed out

insomniacslk:
Hi Wilmer,

I've updated all the room names to match "ELF Arena" with this exact upper/lower case.

Re-reading, I'm not sure where do you have put that `"id": "MOCA 2016"` line - from the comma at the end it seems to be part of `menu.json` but I see no instance of `id` therein. And the ICS has a different format, so better to ask you.

Wilm0r:
Yep, menu.json. right in your entry. Indeed this isn't a common field and I should likely deprecate it as I was thinking out loud in the previous comment.

The idea was that schedule URLs could change so instead the ID was added. But meh. URLs never changed and IDs are mostly missing, in which case giggity falls back to a checksum of the URL.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

insomniacslk:
Thanks Wilmer! I've just reproduced this, and also tested with the ID. The entry starts with "MOCA 2016" and after fetching it it changes into "MOCA, The Italian Hacker Camp", which would be perfectly fine for me. However I've modified the ICS to show just "MOCA 2016" to match the entry in `menu.json`.

Wilm0r:
A full match is nicer of course, yes. :-) Merged.

I hope I'll manage to do another release before the event starts (with the Italian translation). But the menu.json change should go live immediately.

insomniacslk:
Awesome, thanks!

I have just tried switching my phone to Italian to try the translation and it seems to be quite right. As a feedback for the future, what about making the translation accessible via settings (as well as via system language)?

insomniacslk:
Also, any plan for when the updated app will be on the Play Store? I don't want to put pressure on you though, there are still nearly two weeks before the event starts. Thanks!

Wilm0r:
Right now the Italian translation would be the only addition, I'm trying to get another change in (support for reading custom menu.json files/entries so next time someone would do a PR like this one can test it more easily) but if that doesn't happen within the next week I'll just do a translation-only release, no problem.

insomniacslk:
Oh, sorry, my misunderstanding then. I thought that in order to have the MOCA schedule, an app update was required. I just uninstalled and reinstalled and I see the event, so I guess the menu.json is fetched from the network.

In other words, don't worry if the only update is the translation, I was just caring about the menu.json update :)

PR #3, opened 2016-08-05
Added support for JSON files

siddhantbajaj:
added a JSON parser for schedules available in JSON format.


app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -298,7 +303,11 @@ public void loadSchedule(String url_, Fetcher.Source source) throws IOException
 				loadDeox(in);
 			} else if (head.contains("begin:vcalendar")) {
 				loadIcal(in);
-			} else {
+			}
+			  else if(head.contains("{")&&head.contains("[")){
```
Wilm0r: On top of this in theory this could rely on MIME types (mostly not possible elsewhere because it's all XML). But not terribly important I guess. I dislike how fuzzy this is but I've never really been able to do better anyway.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -355,7 +365,68 @@ private void loadXml(BufferedReader in, ContentHandler parser) {
...
+				if ((line = tentMap.get(location)) == null) {
+					line = new Schedule.Line(location, location);
+					tents.add(line);
+					tentMap.put(location, line);
+				}
+				line.addItem(item);
+			}
+
+		} catch (JSONException e) {
+			e.printStackTrace();
```
Wilm0r: Not that Giggity has flawless error handling/reporting as it is, please do make sure this one gets reported to the user. :-)

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -355,7 +365,68 @@ private void loadXml(BufferedReader in, ContentHandler parser) {
...
+		Time=DateTime[1].split(":");
+		int year=Integer.parseInt(Date[0]);
+		int month=Integer.parseInt(Date[1]);
+		int date=Integer.parseInt(Date[2]);
+		int hour=Integer.parseInt(Time[0]);
+		int min=Integer.parseInt(Time[1]);
+		int sec=Integer.parseInt(Time[2]);
+		calender.set(year,month,date,hour,min,sec);
+		return  calender.getTimeInMillis();
+	}
```
Wilm0r: Hrm, I *think* I had to do something quite similar in the Pentabarf parser already and IIRC it wasn't that complicated. Does Java not have a nicer way to do this?

(Also I saw OpenEvent has native timezone handling which is rather nice. Is that dealt with properly here? (Question is actually: what is properly :-))

Wilm0r:
Thanks for this! I've sent a few comments. Could you please also update the style to match the existing code?

Wilm0r:
This one is still open ... are you able to update it?

What events are currently using this file format?

PR #4, opened 2016-12-30
Allow to add events to calendar

fid-jose:
This commit adds a new button which adds the current event to the calendar

### DIFF
--- a/app/src/main/java/net/gaast/giggity/EventDialog.java
+++ b/app/src/main/java/net/gaast/giggity/EventDialog.java
@@ -38,6 +38,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewTreeObserver;
+import android.widget.Button;
 import android.widget.CheckBox;
 import android.widget.CompoundButton;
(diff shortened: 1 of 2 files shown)


Wilm0r:
Sorry for the long delay. :-( Finally figured out how to easily test a PR locally without having to create a separate project in Android Studio. Few things:

For consistency and space reasons, I'd definitely want this to be an icon, not a text button. Though I can't really find a super suitable one among either Android Drawables or the Material Design icon set.. :-/ But maybe one of the icons when you search https://material.io/icons/ for the word "event"?

And what kinda sucks in Giggity and will bite especially here: Timezone awareness. When I wrote Giggity, I think the Pentabarf format didn't have TZ info yet, so I just parse everything as if it were local, and flush all the parsed data on timezone changes so that a reload will happen.

But if calendar events get created, they *will* be for the wrong time if the user is not yet in the timezone of the event... I'd rather not add this feature yet as long as it'd be so sensitive to this bug, what do you think?

Wilm0r:
OK, I may have a plan for making Giggity timezone-aware, now I just need some time to try it out..

PR #5, opened 2016-12-30
Use internal cache if external cache fails

fid-jose:
This commits checks whether it's possible to write in the external cache and if not, it uses the internal cache.

### DIFF
--- a/app/src/main/java/net/gaast/giggity/Fetcher.java
+++ b/app/src/main/java/net/gaast/giggity/Fetcher.java
@@ -114,6 +114,9 @@ public class Fetcher {
 				}
 
 				fntmp = new File(app.getExternalCacheDir(), "tmp." + Schedule.hashify(url));
+				if (!fntmp.canWrite()) {
+					fntmp = new File(app.getCacheDir(),  "tmp." + Schedule.hashify(url));
+				}
 				OutputStream copy = new FileOutputStream(fntmp);
 				inStream = new TeeInputStream(inStream, copy, true);  // true == close copy on close
 


Wilm0r:
In what situation would that write fail by the way?

Seems reasonable, though shouldn't the cache-read code then also try both paths?

fid-jose:
> In what situation would that write fail by the way?

I have a Huawei G Play mini with Android 5.1.1 and without the patch, giggity fails, applying the patch it works well

> Seems reasonable, though shouldn't the cache-read code then also try both paths?

Umm, didn't think about that. I'll have a look at that.

Wilm0r:
Interesting, thank you for finding this and submitting a fix!

I've recently switched to using the external cache because you can download PDFs/etc to it and then invoke an external viewer to show them, handy for conference info booklets, maps, etc. Other stuff could just remain in the internal cache.

So with this problem on your phone I guess that functionality couldn't be made to work without trying out other storage tricks... :-(

Wilm0r:
Interestingly, this trick seems to not work all that well in very recent Android as I've just discovered on my new Pixel:

12-31 15:00:44.205  2184  2184 E AndroidRuntime: android.os.FileUriExposedException: file:/storage/emulated/0/Android/data/net.gaast.giggity/cache/sched.6f48066ecff6e9d69aa74879efa92dacab0a0f5e exposed beyond app through Intent.getData()

I should switch to content:/// URLs somehow.

This Fetcher I'd love to get rid of actually if I could find something similar already built into the Android framework...

Wilm0r:
Can you check whether this works for you? https://github.com/Wilm0r/giggity/commit/eb9e103009d324f9c15e68b405ecd5fc3752991b

Should solve your problem by no longer using external cache, yet it should work with PDF/image viewing by using content:// URLs. (Mind checking that by opening PyconES and then the Info link in the nav drawer?)

Wilm0r:
Closing this one. Sorry, I need to try out your other change, haven't managed to do so yet but not ignoring it!

Please do let me know if you're still having problems even with this change. I've tested it on a few (virtual) devices and it's all good from what I can tell.

fid-jose:
> Can you check whether this works for you? eb9e103

I cannot try it today but I'll try to test it tomorrow

fid-jose:
> Can you check whether this works for you? eb9e103

Applying eb9e103 no longer fails

PR #7, opened 2017-01-10
Make the tent name column wider

maks:
This makes UI bit nicer when tent names are reasonable length.

Ideally the width should be based on the longest tent name string length (with a max limit) but for now this quick fix improves the UI substantially for longer names (eg. LCA2017).


Wilm0r:
Though this definitely looks nicer, it does eat into precious screen estate in portrait mode. :-( Guess we can try it for a while. I can see the benefit of this in your schedule yeah. Words wrapped internally is ugly..

Issue #14, opened 2017-06-17
Adding support for FOSSASIA's open event JSON format 


heysadboy:
FOSSASIA's open event JSON format contains information about rooms, icon and other web links which has to be added separately for the formats like pentabarf along with sessions.

I made the changes https://github.com/heysadboy/giggity .

App could be tested for this API end point https://open-event-dev.herokuapp.com/api/v1/events/266?include=sessions,microlocations

See the endpoints's response here http://jsonviewer.stack.hu/#http://open-event-dev.herokuapp.com/api/v1/events/266?include=sessions,microlocations .

I have modified app so it automatically fetches the information of locations, sessions and icon's link from this endpoint itself.

heysadboy:
Sending the pull request to add the support here https://github.com/Wilm0r/giggity/pull/15 @Wilm0r

heysadboy:
@Wilm0r Should we close this?

PR #15, opened 2017-06-17
Adding support for FOSSASIA open event JSON format

heysadboy:
FOSSASIA's open event JSON format contains information about rooms, icon and other web links which has to be added separately for the formats like pentabarf along with sessions.

Please see this issue for reference https://github.com/Wilm0r/giggity/issues/14

App could be tested for this API end point https://open-event-dev.herokuapp.com/api/v1/events/266?include=sessions,microlocations

See the endpoints's response here http://jsonviewer.stack.hu/#http://open-event-dev.herokuapp.com/api/v1/events/266?include=sessions,microlocations .

I have modified app so it automatically fetches the information of locations, sessions and icon's link from this endpoint itself.

### DIFF
--- a/app/src/main/java/net/gaast/giggity/Schedule.java
+++ b/app/src/main/java/net/gaast/giggity/Schedule.java
@@ -428,8 +428,8 @@ public class Schedule {
 	private void loadJson(BufferedReader in) {
 
 		StringBuffer buffer = new StringBuffer();
-		SimpleDateFormat df = new SimpleDateFormat();
-		HashMap<String, Schedule.Line> tentMap = new HashMap<String, Schedule.Line>();
+        SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        HashMap<String, Schedule.Line> tentMap = new HashMap<String, Schedule.Line>();
 		Boolean hasMicrolocs = false;
 		Scanner s = new Scanner(in);
 		HashMap<String, String> locs = new HashMap<>();
@@ -512,15 +512,28 @@ public class Schedule {
 				String startTimeS = event.getString("start_time");
 				String endTimeS = event.getString("end_time");
 				Date startTime, endTime;
-				startTime = new Date(getTimeInMillis(startTimeS));
-				endTime = new Date(getTimeInMillis(endTimeS));
-				Schedule.Item item = new Schedule.Item(uid, title, startTime, endTime);
+
+                startTimeS = startTimeS.replace('T', ' ');
+                if (startTimeS.contains("+")) {
+                    startTimeS = startTimeS.substring(0, startTimeS.lastIndexOf('+'));
+                }
+                startTimeS = startTimeS.substring(0, startTimeS.lastIndexOf('-'));
+
+                endTimeS = endTimeS.replace('T', ' ');
+                if (endTimeS.contains("+")) {
+                    endTimeS = endTimeS.substring(0, endTimeS.lastIndexOf('+'));
+                }
+                endTimeS = endTimeS.substring(0, endTimeS.lastIndexOf('-'));
+
+                startTime = df.parse(startTimeS);
+                endTime = df.parse(endTimeS);
+
+                Schedule.Item item = new Schedule.Item(uid, title, startTime, endTime);
 				item.setDescription(event.getString("long_abstract"));
 
 				if (event.getString("signup_url") != "null") {
 					item.addLink(new Link(event.getString("signup_url")));
 				}
-
 
 				JSONObject microlocation = event.getJSONObject("microlocation");
 				String location = microlocation.getString("name");
@@ -530,7 +543,6 @@ public class Schedule {
 					tents.add(line);
 					tentMap.put(location, line);
 				}
-
 
 				//Getting value (latitude and longitude) from the map by key (name)
 
@@ -561,32 +573,10 @@ public class Schedule {
 		} catch (JSONException e) {
 			e.printStackTrace();
 			throw new LoadException("Parse error: " + e);
-		}
-	}
-
-	private long getTimeInMillis(String timeString) {
-		GregorianCalendar calender = new GregorianCalendar();
-		String[] DateTime = new String[2];
-		String[] Date = new String[3];
-		String[] Time = new String[3];
-		DateTime = timeString.split("T");
-		Date = DateTime[0].split("-");
-		Time = DateTime[1].split(":");
-
-		//To avoid error in open event JSON format
-		if (Time[2].contains("+")) {
-			Time[2] = Time[2].substring(0, Time[2].indexOf('+'));
-		}
-
-		int year = Integer.parseInt(Date[0]);
-		int month = Integer.parseInt(Date[1]);
-		int date = Integer.parseInt(Date[2]);
-		int hour = Integer.parseInt(Time[0]);
-		int min = Integer.parseInt(Time[1]);
-		int sec = Integer.parseInt(Time[2]);
-		calender.set(year, month, date, hour, min, sec);
-		return calender.getTimeInMillis();
-	}
+        } catch (ParseException e) {
+            e.printStackTrace();
+        }
+    }
 
 	/** OOB metadata related to schedule but separately supplied by BitlBee (it's non-standard) gets merged here.
 	  I should see whether I could get support for this kind of data into the Pentabarf format. */
--- a/app/src/main/java/net/gaast/giggity/Schedule.java
+++ b/app/src/main/java/net/gaast/giggity/Schedule.java
@@ -65,6 +65,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.Locale;
+import java.util.Scanner;
 import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -308,7 +309,9 @@ public class Schedule {
 				loadDeox(in);
 			} else if (head.contains("begin:vcalendar")) {
 				loadIcal(in);
-			} else {
+			} else if (head.contains("{")) {
+                loadJson(in);
+            } else {
 				Log.d("head", head);
 				throw new LoadException(app.getString(R.string.format_unknown));
 			}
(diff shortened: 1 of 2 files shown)


app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -308,7 +309,9 @@ public void loadSchedule(String url_, Fetcher.Source source) throws IOException
 				loadDeox(in);
 			} else if (head.contains("begin:vcalendar")) {
 				loadIcal(in);
-			} else {
+			} else if (head.contains("{")) {
+                loadJson(in);
+            } else {
```
Wilm0r: The rest of the file uses tabs, makes the indents look a bit funky. [...] Oh ah, nevermind, I see that the rest of your submit uses tabs, thanks. :-)

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,155 @@ private void loadIcal(BufferedReader in) {
...
+
+				if (latlon != null) {
+					line.setLocation(latlon);
+				}
+
+				line.addItem(item);
+			}
+
+		} catch (JSONException e) {
+			e.printStackTrace();
```
Wilm0r: Should report a load error to the user as well, similar to how the other loaders do it.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,155 @@ private void loadIcal(BufferedReader in) {
...
+				for (i = 0; i < linklist.length(); ++i) {
+					JSONObject link = linklist.getJSONObject(i);
+					Schedule.Link slink = new Link(link.getString("link"), link.getString("name"));
+					slink.setType(link.optString("type", null));
+					links.addLast(slink);
+				}
+
+				//Adding website separately because it is not in social_links
+				if (conference.has("event_url")) {
+					Schedule.Link slink = new Link(conference.getString("event_url"), "Website");
```
Wilm0r: Just wondering, and maybe I'm showing my age here :-D But should website not be first instead of last?

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,155 @@ private void loadIcal(BufferedReader in) {
...
+
+				//Getting value (latitude and longitude) from the map by key (name)
+				String locString = locs.get(line.getTitle());
+				String latitude = locString.substring(0, locString.indexOf(','));
+				String longitude = locString.substring(locString.indexOf(',') + 1);
+
+				//Adding location details here
+				String latlon = null;
+				try {
+					latlon = ("geo:0,0?q=" + longitude + "," +
```
Wilm0r: Ah, shame, did I put URL-encoding of that in a not very reusable place? I'll refactor that later I guess.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,155 @@ private void loadIcal(BufferedReader in) {
...
+				String location = microlocation.getString("name");
+
+				if ((line = tentMap.get(location)) == null) {
+					line = new Schedule.Line(location, location);
+					tents.add(line);
+					tentMap.put(location, line);
+				}
+
+				//Getting value (latitude and longitude) from the map by key (name)
+				String locString = locs.get(line.getTitle());
```
Wilm0r: microlocations are mandatory in your JSON format?

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,155 @@ private void loadIcal(BufferedReader in) {
...
+			buffer.append(s.nextLine());
+
+		}
+
+		String output = buffer.toString();
+		Schedule.Line line = null;
+
+		try {
+
+			JSONObject conference = new JSONObject(output);
```
Wilm0r: Slurping the whole thing into a StringBuffer, then string, then this parsing makes me sad. But it looks like the methods on https://stackoverflow.com/questions/22461663/convert-inputstream-to-jsonobject and https://stackoverflow.com/questions/31217653/how-to-read-and-write-json-using-json-simple-library work with different JSON parsers, not in Android's core API so they would pull in extra compiled-in dependencies?

I may do some experimenting once this is merged. This is not a big deal but it just looks less clean to me.

app/src/main/java/net/gaast/giggity/Schedule.java
heysadboy: I also think that should go that way, was a little sceptical about it. No problem :+1:

app/src/main/java/net/gaast/giggity/Schedule.java
heysadboy: @Wilm0r Microlocations are not mandatory, an organiser can select that option from the dashboard to modify the link. Umm I think I will add condition to look for microlocations and not use functionality if the url is not presenting it.

app/src/main/java/net/gaast/giggity/Schedule.java
heysadboy: Yes @Wilm0r . It uses library named "simple json parser". I try to avoid using external library if things are happening to keep up with the app's size. But it's okay if you feel like it. I will help in making changes with another library once this gets merged.

app/src/main/java/net/gaast/giggity/Schedule.java
heysadboy: Like this Log.w("JSON parsing", " Unable to parse ",e); ? Or I should make a toast for it?

heysadboy: Got it! No need to make toast. :sweat_smile:

heysadboy:
@Wilm0r Okay, so making the required changes in the recent commit. Thanks for reviewing.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -430,7 +430,9 @@ private void loadJson(BufferedReader in) {
 		StringBuffer buffer = new StringBuffer();
 		SimpleDateFormat df = new SimpleDateFormat();
 		HashMap<String, Schedule.Line> tentMap = new HashMap<String, Schedule.Line>();
+        Boolean hasMicrolocs = false;
```
Wilm0r: Yikes, this is now a complete mix of spaces and tabs.. :-(

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -512,39 +521,45 @@ private void loadJson(BufferedReader in) {
...
+                    if (latlon != null) {
+                        line.setLocation(latlon);
+                    }
+                }
 
 				line.addItem(item);
 			}
 
 		} catch (JSONException e) {
+            Log.w("JSON parsing", " Unable to parse ", e);
```
Wilm0r: Log.w isn't really any more helpful than e.printStrackTrace() :-/ See the LoadException thing above. That should make it to the user properly.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -512,39 +521,45 @@ private void loadJson(BufferedReader in) {
...
-				String latlon = null;
-				try {
-					latlon = ("geo:0,0?q=" + longitude + "," +
-							latitude + "(" +
-							URLEncoder.encode(line.getTitle(), "utf-8") + ")");
-				} catch (UnsupportedEncodingException e) {
-					e.printStackTrace();
-				}
+                if(hasMicrolocs) {
+                    String locString = locs.get(line.getTitle());
```
Wilm0r: So the problem is more ... what will locs.get("non-existent-or-misspelled-name") do? https://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html#get(java.lang.Object) looks like it'll return null, which is going to end poorly below.

app/src/main/java/net/gaast/giggity/Schedule.java
Wilm0r: No, I'll take this over pulling in an external library, let's leave it like this then. I'm rather keen on keeping the Giggity .apk very small as it is now. Just a shame that the built-in JSON lib only has a string parser. :-/

app/src/main/java/net/gaast/giggity/Schedule.java
heysadboy: @Wilm0r If think there would be any misspelled name the match would be that misspelled name. It depends on organiser when the event is created on our server, but I think this works fine.
Anyways it'd great if you could suggest any other way.

I have made other changes from the reviews.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,170 @@ private void loadIcal(BufferedReader in) {
...
+				if ((line = tentMap.get(location)) == null) {
+					line = new Schedule.Line(location, location);
+					tents.add(line);
+					tentMap.put(location, line);
+				}
+
+
+				//Getting value (latitude and longitude) from the map by key (name)
+
+				if (hasMicrolocs && line.getTitle()!=null && !line.getTitle().equals("")) {
```
heysadboy: @Wilm0r Adding this null check.

Wilm0r: The one I was more concerned about it the one below.

But anyway, it's all not *that* important if you ensure that the input you provide for your conference is valid. Time's going to run out in a bit and we should allow for some time for people to upgrade to the new Giggity .apk. So I guess we should merge this.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -421,6 +424,170 @@ private void loadIcal(BufferedReader in) {
...
+				line.addItem(item);
+			}
+
+		} catch (JSONException e) {
+			Log.w("JSON parsing", " Unable to parse ", e);
+			e.printStackTrace();
+		}
+	}
+
+	private long getTimeInMillis(String timeString) {
```
Wilm0r: Oh yeah, one thing: Can't you use https://developer.android.com/reference/java/text/SimpleDateFormat.html#parse(java.lang.String,%20java.text.ParsePosition) for this?

heysadboy: Thanks for looking up. I need to make few more changes as our format is being updated. Here you can read more about new format https://open-event-api.herokuapp.com/ , I guess this will be helpful for readme.

heysadboy:
I have made the final changes please see @Wilm0r

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -512,16 +512,29 @@ private void loadJson(BufferedReader in) {
 				String startTimeS = event.getString("start_time");
 				String endTimeS = event.getString("end_time");
 				Date startTime, endTime;
-				startTime = new Date(getTimeInMillis(startTimeS));
-				endTime = new Date(getTimeInMillis(endTimeS));
-				Schedule.Item item = new Schedule.Item(uid, title, startTime, endTime);
+
+                startTimeS = startTimeS.replace('T', ' ');
```
Wilm0r: You're doing this because T in the format string above doesn't work?

But I *think* you can make the format string something like "yyyy-MM-dd'T'HH:mm:ss" so this substitution won't be necessary.

heysadboy: Thanks for reviewing, making the changes.

heysadboy:
Thanks @Wilm0r :-)

PR #17, opened 2017-07-24
Add DebConf17

stefanor:
Preview:
debconf17

The icon doesn't look great on a dark background, but it's what we have...

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -42,6 +42,11 @@
 						"url": "https://wiki.debconf.org/upload/8/88/DebConf17_venue_layout.png",
 						"title": "Map",
 						"type": "image/png"
+					},
+					{
+						"url": "https://debconf17.debconf.org/static/docs/welcome.pdf",
+						"title": "Welcome Guide",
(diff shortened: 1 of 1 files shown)


app/src/main/res/raw/menu.json
```diff
@@ -22,6 +22,31 @@
 			}
 		},
 		{
+			"version": 2017072400,
+			"url": "https://debconf17.debconf.org/schedule/pentabarf.xml",
+			"title": "DebConf17",
+			"start": "2017-08-05",
+			"end": "2017-08-12",
+			"metadata": {
+				"icon": "https://debconf17.debconf.org/static/img/favicon/favicon-180.ea5f30d24599.png",
```
stefanor: Not sure what the preferred resolution is, but that seems big enough for most cases.

Wilm0r:
There's no preferred resolution, it's only important that it's square and has an alpha layer. I think for some uses it has to be monochrome as well but I'm not sure whether I should keep enforcing that ... it may do a decent enough conversion for .... notifications? not sure.

QR I only now see so I tested this the old fashioned way, but all looks good, thank you!

PR #21, opened 2017-08-08
Added French translation

mmuman:
Untested as I don't have the Android toolchain set up here.


Wilm0r:
Should be safe enough, thanks!

(If you do happen to notice some oddness later on when this is released please feel free to send corrections.)

PR #24, opened 2017-09-05
Add PyConFR 2017

mmuman:
Icon is too large and on white bg but I didn't find anything else.

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -434,5 +434,43 @@
 				]
 			}
 		}
+		{
+			"version": 2017090500,
+			"url": "https://cfp.pycon.fr/schedule/xml/",
+			"title": "PyConFR 2017",
+			"start": "2017-09-21",
+			"end": "2017-09-24",
+			"metadata": {
+				"icon": "https://www.pycon.fr/2017/theme/images/icon2017.png",
+				"links": [
+					{
+						"url": "https://www.pycon.fr/2017/",
+						"title": "Site Web"
+					},
+					{
+						"url": "https://www.eventbrite.fr/e/billets-pycon-fr-2017-a-toulouse-37380880219",
+						"title": "Inscription"
+					},
+					{
+						"url": "https://www.pycon.fr/2017/pages/venir.html",
+						"title": "Venir"
+					}
+				],
+				"rooms": [
+					{
+						"name": "A.*",
+						"latlon": [43.60248, 1.45464]
+					},
+					{
+						"name": "B.*",
+						"latlon": [43.60247, 1.45566]
+					},
+					{
+						"name": "C.*",
+						"latlon": [43.60214, 1.45462]
+					}
+				]
+			}
+		}
 	]
 }


mmuman:
Looks like my Android 4 is getting really old though, can't load the schedule with SSL likely due to too recent cyphers and old ones all disabled. Proxying it on http works though.

sblondon:
Checked with Android 7 ; there is no problem with SSL for the XML URL.

sblondon:
@mmuman What size should we have for the logo? A transparent background is better?
In the application, I don't see logos for the others conferences.

heysadboy:
@mmuman @sblondon I tried to remove the background of the image and making it square shaped. Will this work?

logo

Wilm0r:
So on logos ... they're not mandatory. If you don't have a suitable one, then don't worry about it, better to omit than to include an unsuitable one.

The issue here is, I think for some purposes they need to be monochrome (not 1-bit but single chrome with shades or probably really alpha layer), mostly when used as a notification icon. When used as an icon for the home shortcut, it's not as important, mostly it needs to be high-dpi.

But if there's no logo specified it'll just use the Giggity logo. Up to you folks. :-)

(Varun, I'd go for simplified with the two lines of text removed and really just the logo, but ... up to the submitters in the end of course, let's see.)

heysadboy:
@Wilm0r Maybe we could try with the attached image, the choice is of submitters anyways. Edges look little blurry but will look okay when minimized to icon size. 
logo1

mmuman:
Looks ok.
I managed to generate a little less blurry one from the SVG:
``` bash
$ rsvg-convert --zoom 2 logo2017.svg > logo-pycon2017.png
$ convert -crop 240x240+260+20 logo-pycon2017.png icon-pycon2017.png
```

Now we just need it hosted on the website to change the icon URL in the JSON.

heysadboy:
@mmuman That's very great. I tried to remove background too so looking blurry because of that maybe, you can see transparent background when you open the image in new tab.

mmuman:
Well the SVG is transparent, and rsvg and convert preserve this so just do it

heysadboy:
Oh, that's good @mmuman! Didn't know about that.

sblondon:
This icon is available at https://www.pycon.fr/2017/theme/images/icon2017.png

@mmuman Can you update your pull-request?

mmuman:
Cool thanks!

mmuman:
I believe we can merge this one now ;-)

Wilm0r:
This is manually merged now with a few corrections (missing , and updated main version# which admittedly is a bug in the menu.json handler that I should just fix).

Merged by hand because Git is shit. :-/

mmuman:
Oh right, now I see the missing comma. It's hard to test without being able to build it.
Gedit highlight extra chars but the missing one went unnoticed.
Now, I can test the syntax with json_pp:
```shell
$ json_pp -f json -t json < app/src/main/res/raw/menu.json > /dev/null
```

mmuman:
Hmm, it should be possible to add a git commit hook to reject commits with boggus json...

mmuman:
Yeah I just added the json_pp test in my `.git/hooks/pre-commit` and it works :-)

sblondon:
Thank you all ! :-)

Issue #26, opened 2017-09-21
DURATION not supported in ICS ("Schedule appears to be empty" on the Paris Web 2017 ICS)


mmuman:
Trying to load the [Paris Web 2017 ICS](http://www.paris-web.fr/2017/conferences-parisweb-2017.ics) takes a lot of time, then :
"Schedule appears to be empty"

mmuman:
At least it's valid UTF-8…
Ohhh, got it : it doesn't have DTEND but instead events have a DURATION. First time I see this.

mmuman:
So, managed to write a python script to massage it until Giggity gets a fix:
```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import re
import time

content = ''

with open("conferences-parisweb-2017.ics", 'rb') as f:
	dtstart = ''
	for line in f:
		dline = line.decode()
		# cleanup bad escaping
		dline = dline.replace('\\\\\\,', ',')
		dline = dline.replace('\\\\,', ',')
		dline = dline.replace('\\,', ',')
		# convert DURATION to DTEND
		if dline.find('DTSTART:') == 0:
			dtstart = dline
		if dline.find('DURATION:') == 0:
			m = re.match('^DURATION:PT(.*)M.*$', dline)
			duration = int(m.group(1))
			tm = time.strptime(dtstart, 'DTSTART:%Y%m%dT%H%M%SZ\r')
			tm = time.mktime(tm)
			tm = tm + duration * 60
			tm = time.localtime(tm)
			tm = time.strftime('DTEND:%Y%m%dT%H%M%SZ\r', tm)
			print(dtstart)
			dline = tm
			print(m.groups())
			print(dline)
		content = content + dline

with open("conferences-parisweb-2017.fixed.ics", 'w') as f:
	f.write(content)
```

mmuman:
Of course they changed their ICS the day before the event, but without adding a DTEND. The fixed python (and I missed a \n previously anyway):
```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import re
import time

content = ''

with open("conferences-parisweb-2017.ics", 'rb') as f:
	dtstart = ''
	for line in f:
		dline = line.decode()
		# cleanup bad escaping
		dline = dline.replace('\\\\,', ',')
		#dline = dline.replace('\\,', ',')
		# convert DURATION to DTEND
		if dline.find('DTSTART;VALUE=DATE-TIME:') == 0:
			dtstart = dline
[...]

PR #27, opened 2017-09-21
WIP: Add Paris Web 2017

mmuman:
DO NOT MERGE YET:
WIP: the ICS "appears to be empty" for some reason...

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -471,6 +471,27 @@
 					}
 				]
 			}
+		},
+		{
+			"version": 2017100400,
+			"url": "http://revolf.free.fr/conferences-parisweb-2017.fixed.ics",
+			"title": "Paris Web 2017",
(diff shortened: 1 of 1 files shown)


mmuman:
Hmm conference is in two days anyway. Still, would be nice to know the why of #26 ...

mmuman:
Ok, now we just need them to host a fixed version of the ICS.

mmuman:
I've put a fixed version on my own website for now, I don't know if they'll have the time tonight to host it...

For the record, I've added `--json_opt utf8,pretty` to the test in my `pre-commit` hook.

Wilm0r:
OK, and now it's already over so no point in merging? :-( Sorry about that. But did you manage to get it used by for example hosting a gzip-QR version of your json blob?

mmuman:
Yeah I posted the QR on twitter & mastodon, not sure it was of any use. I think we can close it now. We can always dig that next year.
We should still try to add support for DURATION in ICS files.

PR #28, opened 2017-09-21
WIP: extract ics speakers

mmuman:
Some attempts at extracting speaker name(s) (currently only one) from various ICS.

Totally UNTESTED as I still can't build the app with gradle here, it doesn't even attempt to actually compile the java code so I have no way of making sure it's valid syntax.

Please test.


mmuman:
pattern would be something like `(?<name>.*) — (?<speaker>.*)` for Paris Web, modulo proper escaping.

mmuman:
Hmm I think role=req-participant is the default, according to spec, so maybe we should only avoid other values.

mmuman:
Ohhh, I think I finally found out how to build on Debian Sid with the packages. Had to install some google-android-* stuff, including google-android-m2repository-installer.
Now I need to fix the java syntax!

mmuman:
"added in API level 26" for Matches::group(String) :^)

mmuman:
I still can't link the app but at least the java syntax should be ok now. I reworked the last commit to use an unnamed group in the regex. Still untested.

Wilm0r:
OK, interesting. So just to clarify for myself (sorry, bit late for me to try to parse this all from the Github code review UI :-/ ) : 

One would add a "speakermatch" attribute to the metadata for the conf (so that will only work for things in menu.json, or actually also if a (gzipped) JSON blob QR is put somewhere online for example!), which will instruct the parser to split talk descriptions cleverly into name of speaker and actual name of talk?

But I separately see mentions of the (req-)Participants fields? Which would be the preferable way, but "speakermatch" is the fallback when speaker name isn't properly separated in metadata?

mmuman:
Yes, that's two ways of doing it, either by adding a contact as participant, or by munging the summary field.

I tried to add the author as participant for an event I work on, but Thunderbird keeps adding myself as organizer for some reason, and NextClound on the other hand accepts it but the whole field is just missing in the ICS export, so I'll have to use the summary as well for the Alchimie.

Wilm0r:
:-(

Yeah really xcal is just not a suitable format for this. In a way I kinda want to discourage its use, instead of jumping through lots of hoops of heuristics. Guess we can at least take this one, though? But have you been able to build/test this?

mmuman:
The code compiles but I can't get the build finished, so I can't test it.

app/src/main/java/net/gaast/giggity/Schedule.java
```diff
@@ -948,6 +969,11 @@ public void startElement(String uri, String localName, String qName,
 			if (localName.equals("vevent")) {
 				eventData = new HashMap<>();
 			}
+			if (localName.equals("attendee")) {
+				if (atts.getValue("role") == "REQ-PARTICIPANT")
+					eventData.put("speaker", new String(atts.getValue("cn")));
+					// TODO: handle more than one?
```
snoack: Yes, that is for example how multiple speakers are provided in FOSDEM'S iCal file.

PR #35, opened 2017-11-01
adding linuxday.at/2017 to menu.json

ka7:
adding linuxday.at/2017
2.dec.2017

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -628,12 +628,8 @@
 						"url": "https://www.linuxday.at/workshops/2017",
 						"title": "Workshops"
 					}
-
-
-
 				]
 			}
 		}		
-
 	]
 }
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -602,6 +602,38 @@
 					}
 				]
 			}
-		}
+		},
+		{
+			"version": 2017110100,
+			"url": "https://www.linuxday.at/vortraege/2017/calendar.ics",
+			"title": "LinuxDay.at 2017",
+			"id": "linuxday_at_2017",
+			"start": "2017-12-02",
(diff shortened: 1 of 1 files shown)


mmuman:
Hi, thanks for the contribution!

Some remarks:
- there are some whitespaces (tabs and empty line) at the end of the block you added, could you clean them up?
- the ICS seems to lack a X-WR-CALNAME property, and Giggity doesn't like that much, it displays the full ICS URL instead. Do you have any contact with the organizers to have it changed?

ka7:
Fixed the blank lines - lets win the beauty contest. :)

x-wr-calname: generated by Drupal, found this interesting link: https://www.drupal.org/node/1887572; anyway, we activated it now.

Btw: i run myself some local tests (jslint, validate each oft my URL) to ensure good data.
Could be a job for travis-ci ? ( checks, testcompile, ...) - @fossasia worked already into this direction: https://github.com/fossasia/giggity/blob/master/.travis.yml

mmuman:
Oh, I didn't know CALNAME made Outlook to go nuts, good to know...

Indeed having CI would be nice, but only Wilmer can set this up I think.

Wilm0r:
mmuman, thank you so much for jumping on top of this so I can immediately merge. :-D

And yes ,that id bug needs to go. Initially I implemented id's because I hated the idea of using URLs as primary keys, but ... this has its problems as well or at least the current implementation does.

CI could give me some sort of presubmit? That would be quite neat. Though I am also wondering whether I can have a different mechanism from PR's for accepting new menu.json entries..

Wilm0r:
Done https://github.com/Wilm0r/giggity/commit/16ca9f3dd00733a2640a67da34bdfec4c61c7691

mmuman:
Hmm, the id is incorrect, it should be "LinuxDay Vortraege".

PR #37, opened 2017-11-07
fixing the ID thing

ka7:
.. fixing the ID thing.
(think so.. ; also will update it in the .ics)

there are multiple similar LUGS & events
linuxdayS.cz - this is in Czech
linxuday.at - this is in austria ( this )
linxuday.it - italy based..


Wilm0r:
Done

PR #39, opened 2017-11-08
Add the bump_menu_rev shell tool

mmuman:
It just updates the menu revision with the current date, and creates a
commit for this.

It doesn't check if the actual counter needs to be incremented if date
is the same, but it doesn't seem to be needed yet.

### DIFF
--- a/tools/bump_menu_rev.sh
+++ b/tools/bump_menu_rev.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+sed -e "2s/\(^\t\"version\": \).*$/\1$(date '+%Y%m%d01'),/;" -i app/src/main/res/raw/menu.json && git add app/src/main/res/raw/menu.json && git commit -m "Bump menu revision"


Wilm0r:
Hrm, I'd rather leave this with you as a personal utility. I'd normally not want a separate git revision just with a version number bump, for example..

Wilm0r:
Closing as this one's (IMHO) too specific to your flow.

mmuman:
I still don't really understand why you absolutely want less commits, I mean commits are quite cheap. This goes both against the normal github workflow and the concurrent submissions of multiple patches.
Just see the PRs for FOSSASIA & Newline, they both increment the global revision, so you can't just merge them as is.

IMO the JSON revision itself should be tied to the publication, not the commits adding talks, hence in a separate commit just before putting it online. Or do you just redirect to the raw content on github? In which case it can be a problem having inconsistent data before the revision is bumped. Otherwise I don't see any problem here. And even in this case, it could be easily solved with a separate "menu-release" branch which is updated when needed. 
The current workflow also implies manual merging which steals your time and could be avoided…

PR #44, opened 2017-12-26
Also share speakers and weblinks

t-8ch:
This includes speakers and a weblink in the shared information.
The weblink is a newly introduced concept as is supposed to represent the URL where the event is accessible on the web.
I would have prefered to call this `URL` but the `url` field already fulfills a different role.

This new field is only parsed from Pentabarf files, but should be generalizable.
(I am willing to implement it for the other formats)


Wilm0r:
Sorry for the delay! :-(

PR #45, opened 2017-12-31
Add FOSDEM 2018

mmuman:
Same as previous year, just added a link to the stands.

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -709,6 +709,72 @@
 			"start" : "2017-12-26",
 			"end" : "2017-12-30",
 			"version" : 2017121700
+		},
+		{
+			"version": 2018010100,
+			"url": "https://fosdem.org/2018/schedule/xml",
+			"title": "FOSDEM 2018",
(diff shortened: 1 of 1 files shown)


Wilm0r:
Thanks, yeah, funnily I was working on this a few hours before you sent in this PR already :-) But noticed that the info booklet on the FOSDEM site is still last year's. :-(

I have vague memories of refreshing of metadata not working reliably so I need to check that first and obviously fix it, in the meantime trying to avoid getting partially stale data in the seed...

mmuman:
I asked on twitter & IRC, let's see…

Wilm0r:
Confirmed that the refresh does work as it should, so I've pushed this out now. Pfew. :-)

Wilm0r:
FWIW the booklet appeared and is added in https://github.com/Wilm0r/giggity/commit/76ed9e9feb8ea67d67adc3fa5c1ef1faaef87591

Still wondering whether I'm going to add support for room availability status info at the last moment https://api.fosdem.org/roomstatus/v1/listrooms

PR #48, opened 2018-03-03
WIP: Add EclipseCon France 2018

mmuman:
ICS for 2018 isn't online yet.


Wilm0r:
Yay, the first PR with the Travis bits in place :-)

So https://travis-ci.org/Wilm0r/giggity/jobs/348642574 has some warnings on insecure HTTPS fetches that I'll need to address - the content is definitely fine from what I can tell, I just checked with my browser.

But the .ics still just gives a login page which is probably what you were referring to with "not online yet"?

The 32×32 icon is probably really not very useful though on today's high-dpi phone displays?

mmuman:
Yes I used the 2017 one for testing.
The only issue is the calname which is really large, hopefully they'll fix it this year.
I didn't find anything else as icon.

Wilm0r:
OK, so we should wait for the 2018 one to appear before merging?

mmuman:
Yes, we could have some surprises if they changed the format from last year…

mmuman:
Looks like they published the schedule, seems to work fine. I added some more rooms, and dropped the icon.

PR #49, opened 2018-03-09
Add FOSSASIA 2018

mmuman:
Looks like we'll only get an ICS this year… but at least it works fine.


mmuman:
I have the JDLL to add as well, do you prefer a single PR you can squash ?

mmuman:
Just rebased, but I had to fix the conflict on the version field…

mmuman:
would be nice to get this one in, it just started…

PR #59, opened 2018-05-03
short-term (starting tomorrow): Add Linuxwochen Wien 2018

chrysn:
Linuxwochen Wien (a 3-day FLOSS event in Vienna with 3 tracks plus workshops and Python side events) starts tomorrow (or actually today); if that could make it in it'd be great, but obviously I am very very late, and I'll perfectly understand rejection of this PR.

CI checked and did not complain. The event has no square logo I could easily find.

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -1,5 +1,5 @@
 {
-	"version": 2018032101,
+	"version": 2018050301,
 	"schedules": [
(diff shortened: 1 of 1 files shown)


Wilm0r:
Sorry, yeah, you caught me on or near an intercontinental flight at the start of a lots-of-offline trip. Little point in merging right now. :-(

PR #65, opened 2018-06-07
Linux event tuebix 2018

christophprokop:
now with tabs :)

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -1,26 +1,6 @@
 {
 	"version": 2018060701,
-	"schedules": [		
-		{
-			"version": 2018060500,
-			"url": "https://tuebix.org/2018/giggity.xml",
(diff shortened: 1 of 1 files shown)


Wilm0r:
Hrmm, thank you for reminding me to add a new test to the CI when I have 
a chance (-: mind moving this to the bottom to keep the file 
chronological?

(though now I wonder, should it be sorted by conf start date? It may 
just be sorted by addition date ATM which is harder to test properly 
maybe)

christophprokop:
thx for your fast response!
now at the bottom of the file :)

christophprokop:
our linux event is on the day after tomorrow, so merging today would be cool ;)

Wilm0r:
Merged ... but it looks like the &# escaping of ü was unnecessary and caused the meny entry to look ugly until the file is actually loaded (at which point the menu.json title gets overwritten with data from your schedule).

PR #66, opened 2018-06-21
Update GUADEC information for 2018

benzea:
GUADEC is happening soon, would be nice to have this years schedule right in the list.


Wilm0r:
I kinda prefer not editing in place but adding a new entry at the bottom but ... merged to save time.

benzea:
Thanks! Sorry, was not aware of that policy.

PR #67, opened 2018-07-04
Add RMLL 2018

mmuman:


app/src/main/res/raw/menu.json
```diff
@@ -688,6 +688,56 @@
...
+						"title": "Website"
+					},
+					{
+						"url": "https://2018.rmll.info/fr/program/village",
+						"title": "Village du Libre"
+					}
+				],
+				"rooms": [
+					{
+						"name": "A[tT][rR][iI][uU][mM].*",
```
Wilm0r: Ah oh dear, what kind of inconsistency in the schedule is that? :-( I could just make the regex match case insensitive, in fact I'm surprised I haven't already..

PR #75, opened 2018-09-27
linuxday.at @2018-10-13

ka7:
hope the format is ok.


Wilm0r:
Very sorry for having missed this :-(

Merging these, especially with a few in the queue, is a mess. I'm thinking of ways to improve this so this can all hopefully become somewhat automatic.

PR #82, opened 2018-11-28
WIP: Add JM2L

mmuman:
Not sure of the room positions yet.
The ICS misses an X-WR-CALNAME for now.


mmuman:
Sadly the event got canceled, hopefully next year will get more speakers.

PR #84, opened 2018-12-18
Enforce python 3.6 early in CI tests

insomniacslk:
The script `ci/menu-ci.py` requires features only available in Python 3.6, which is not available by default on every distributions (e.g. in Debian it's part of `testing`).

This could be changed to allow lower version numbers, but until then it would be good to break early so the user won't have to look up what's wrong from the exception, and figure out themselves that the Python version is not recent enough.

I am not changing the hashbang to `#!/usr/bin/env python3.6` because that would also stop from using newer versions

### DIFF
--- a/ci/menu-ci.py
+++ b/ci/menu-ci.py
@@ -14,12 +14,21 @@ Sadly I'll need to duplicate some things from Giggity's Java code I guess...
 
 
 import json
-import jsonschema
 import os
-import PIL.Image
 import re
 import subprocess
+import sys
+
+# version check. This script requires at least python 3.6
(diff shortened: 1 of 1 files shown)


insomniacslk:
One issue that I see: if the user has a wrong python version AND missing libraries (like PIL or jsonschema), the version check won't fail before the user has unnecessarily installed all the dependencies.
The "fix" would be to import all the standard libraries first, do the version check, and only then import the non-standard libraries. I am not a big fan of mixing code and imports. What's your advice here?

insomniacslk:
Updated as per offline discussion with @Wilm0r

PR #86, opened 2019-01-18
Add Linux.conf.au 2019

pserwylo:
This also archives the old "Linux.conf.au 2018" to the correct URL's. Currently they are pointing to the linux.conf.au host whereas they should point to 2018.linux.conf.au. The new linux.conf.au 2019 stuff could equally point to linux.conf.au, but it works even better with 2019.linux.conf.au, because in the future there will of course be a 2020 conference.

### DIFF
--- a/app/src/main/res/raw/menu.json
+++ b/app/src/main/res/raw/menu.json
@@ -1,5 +1,5 @@
 {
-	"version": 2019011600,
+	"version": 2019011800,
 	"schedules": [
 		{
 			"version": 2017041700,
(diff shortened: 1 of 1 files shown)


Wilm0r:
I think the icon will work okay as a home screen icon this way, but not so great in for example notifications, where I think I am using it as well. Let's ... just see how it works out. Just got a reminder that LCA is happening already.

FWIW normally keeping the old edition in the menu isn't really useful, but I'm hoping to soon replace this monolithic file with a bunch of separate json's so let's just take care of the tidying issue that way.

PR #90, opened 2019-02-25
WIP: Add FLOSSCon 2020

mmuman:
Turns out FLOSSCon uses OSEM which actually generates an xml file we can
use. Noticed this too late for 2019, so let's make sure we won't miss
the next one.


Wilm0r:
Oh, those dates are in the past now, and the file actually 404s.

mmuman:
Actually it seems it didn't happen but they scheduled another even in March, likely smaller:
https://www.flosscon.org/conferences/uFLOSSCon2020

Wilm0r:
Ah! Cool, yes. please do :-)

PR #91, opened 2019-02-25
Add JDLL 2019

mmuman:
Not sure the ICS will work again, time will tell.

### DIFF
--- a/menu/jdll_2019.json
+++ b/menu/jdll_2019.json
@@ -1,0 +1,20 @@
+{
+	"version": 2019032300,
+	"url": "https://pretalx.jdll.org/jdll2019/schedule.xml",
+	"title": "Journées du Logiciel Libre 2019",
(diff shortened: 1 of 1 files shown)


mmuman:
This year they switched to Pretalx for the CfP, which is able to generate a pentabarf compatible XML file, so hopefully I'll get an URL to test soon.

mmuman:
Hmm, added the Mastodon avatar as icon, not transparent but it's square at least. Can't really test it as it never worked on my Android 4 tablet.
Also their SSL setup makes Android 4 freak out, so I had to copy the xml around for testing.

mmuman:
ping :-)

Wilm0r:
Sorry for the delay! Missed that this wasn't draft anymore :-(

mmuman:
Thanks!

PR #93, opened 2019-02-25
Add LIT 2019

iblech:
The LIT was included in Giggity for the 2017 installment. In 2018, we unfortunately were too late in submitting the pull request. We hope that it'll work again for 2019. :-)

The linked schedule is not yet accessible, but will be soon.

Thank you for your efforts in maintaining Giggity. :-)

### DIFF
--- a/menu/lit_2019.json
+++ b/menu/lit_2019.json
@@ -1,0 +1,16 @@
+{
+	"version": 2019022500,
+	"url": "https://frab.luga.de/de/LIT2019/public/schedule.xml",
+	"title": "Augsburger Linux-Infotag 2019",
+	"start": "2019-04-06",
+	"end": "2019-04-06",
+	"metadata": {
+		"icon": "https://luga-ev.github.io/frab-extras/Tux2.png",
+		"links": [
+			{
+				"url": "https://www.luga.de/Aktionen/LIT-2019/",
+				"title": "Website"
+			}
+		]
+	}
+}


Wilm0r:
Cool, yes, with 5-6 weeks notice, surely we can get this merged in time now, sorry for missing it last year! When do you think you'll have the schedule file ready?

iblech:
In approximately three weeks. Would it be better if we would publish a placeholder schedule now, to avoid errors?

Wilm0r:
Hrm, a placeholder would be slightly better than nothing at all.

3w from now is only about a week before your conference right? Yeah it's probably better to merge this now in case I'd not pay attn in time that week.

Wilm0r:
Got a schedule file yet?

iblech:
Dear @Wilm0r, the schedule is now online! From my point of view, the pull request can be merged. :-)

Wilm0r:
Oops, realised your serial# was too old so I've patched that in menu.json by hand and ... I guess some day I'll need to figure out why CI failed to catch that.

iblech:
Awesome, thank you for the quick merge!

What do you mean by `serial#`, do you mean the number in the `version` field? In any case, I have very little time on my hands, but I do have a special fondness for debugging Travis issues, so I might be able to have a look.

Wilm0r:
np :-)

The issue is purely with the ci/meni-ci.py script, not with Travis integration. It just misses a check, or I've broken the check during the recent transition away from the monolithic checked in menu.json file, probably the latter. :-(

PR #94, opened 2019-02-27
Add Installfest 2019 to menu

jakubjancicka:
I know we've sent PR at the last moment, but be so kind and merge json for Installfest. It's Czech conference about Linux and open-source technologies which takes place on March 2–3. Thank you (and thank you for giggity, it's great app).


Wilm0r:
Although doing an .apk release is going to be very tight (don't think I'll be able before the weekend, I'll be honest), would you like adding a Czech strings translation for completeness (and maybe for a subsequent year)?

PR #99, opened 2019-04-19
Add opensouthcode 2019 schedule

davidsf:


mmuman:
Hi, thanks for the PR.

We now generate the global menu from a set of json files in menu/ at the root of the project, to simplify merging.
Please put it in a separate file there, see other PRs.

davidsf:
I have put in his own file. I need to change the version of menu.json?

Wilm0r:
No. If the test is still failing, sorry that's my bug. You can fix that by bumping up your version to ...2601 but as said this is a bug, if the CI is happy otherwise then don't worry.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mmuman:
You should be able to `git rebase -i --autosquash master` to get rid of the commits on menu.json and have only the separate file commit left. Then you can `push -f` for force-update your branch here.

mmuman:
And you can even `git commit --amend` to modify the commit log if you want.
Those things should not be done on published history, but on pull-requests it's usually ok, and it ensures the history is readable when merged.
Thanks!

PR #103, opened 2019-05-05
PSES 2019

aeris:


Wilm0r:
Hrm, the issue pointed out by he check is still ongoing? live.passageenseine.fr not resolving here.

aeris:
live will be our streaming, but not operating at this moment.

PR #104, opened 2019-05-06
Add RML13 (public part)

mmuman:
The event has two parts, one for devs only and one for the general
public, and they decided to split the website, and schedule…

Still missing some places.


mmuman:
Didn't have time to add the other places anyway…

Wilm0r:
:-( Sorry for missing this, and the event is over now. There's another PR for RML 13 BTW?

mmuman:
Yeah, at one point they had all on a single website with a proper ICS, then they split the public and the developer events on two websites, one with an ICS, and the other without… go figure. Hopefully it'll be less messy next time.

PR #108, opened 2019-06-09
Update German translation of strings.xml

zuphilip:


Wilm0r:
Great, thank you! Hadn't realised this one had gone so stale. :-(

zuphilip:
Are the other files in that folder also in use? They also differ...

Wilm0r:
You can ignore that, it was dumb duplication which I'm cleaning up in
the quagmire branch: https://github.com/Wilm0r/giggity/blob/quagmire/app/src/main/res/values-de/arrays.xml

PR #110, opened 2019-07-08
Add Kieler Open Source und Linux Tage 2019

Moini:
Hi :)

Our schedule isn't available yet, so I will let this lie until ~ July 30th or so.

How much time does it usually take from merging to the updated app being available on fdroid? (i.e. how much time in advance does this need to be finished)

We will also host another week of events, with one week in between, which is related (Digitale Woche Kiel im Kitz, Sept. 7th-14th). 

What would happen if those events' dates were included in the ics file? Would it break things for Giggity?

Thank you!


Wilm0r:
Hello!

The seed menu file is a json file fetched by Giggity once every 24h at startup time, nothing goes via F-Droid. Although that would in ways be nicer, it'd also be *much* more overhead if there aren't also some relevant code changes to share.

I'd certainly prefer separate files for the two events. Giggity does support gaps (ie empty days) in schedules and will omit them from the date selector for example, but it still seems a bit ugly and counter-intuitive, no?

Sorry for the late response. Note that I'm not always very fast at merging stuff and will indeed be hosting visitors at home (meaning very little computer time) near the end of the month, sorry :-/

Moini:
Hi @Wilm0r and thank you for the explanations. 

So Giggity gets the event info from here, if I understand correctly. That makes it a lot easier and faster, I agree :)

I'll make a second file then, and will hope you'll find some time to merge between the 30th and the 7th for it. If not, I can still give people the link to the ics file and they can add it to their Giggity themselves.

Moini:
Hi again @Wilm0r - the ics file for kielux is online now. I'll make another PR for the Digitale Woche events. 

The days in that one are a bit disconnected, so you need to decide if you would like to take them up then or not.

Moini:
@Wilm0r Thank you!

PR #111, opened 2019-07-14
Add DebConf19 to the menu

stefanor:
https://debconf19.debconf.org/


Wilm0r:
Hrmm, wondering whether CI is broken or otherwise how your change passed it ... since the version# is still in 2018. Could you please correct that, at least?

stefanor:
The content was written in 2018 :P

PR #114, opened 2019-08-01
Add PyCon AU 2019 to menu

auscompgeek:
PyCon AU has started!

For those who don't want to wait until this merged, here's a QR code:

pyconau_2019


auscompgeek:
The conference is over now, so not much point in adding this anymore I guess.

PR #115, opened 2019-08-02
Entry for Digitale Woche Kiel im Kitz (organized by kielux)

Moini:
Hi, @Wilm0r - as promised, here comes the separate entry for the other chain of events.


Moini:
Thank you very much, @Wilm0r !

PR #120, opened 2019-08-11
state of the map: latlon swap

joker234:
Sorry for the little issue :see_no_evil:

Thx for your work :+1: 

Btw. why did I have to increase the version number with a trailing one? Shouldn't `2019081100` be enough?


joker234:
This time I used your advice in the README and checked the Locations before submitting the PR.

Wilm0r:
Oops, nice swap :-)

The version# is for the merged file so if the check wants it higher that means another file already had a higher version#. Not entirely logical yeah, I understand.

PR #117, opened 2019-08-06
Added State of the Map 2019

joker234:
`ci/menu-ci.py` worked. Hope you can merge this PR :)

PR #121, opened 2019-08-21
use unified schedule for State of the Map

tyrasd:
Previously (#117), the SotM program xml was split between a general and an academic track. Thanks to @zverik We now have a unified xml file containing all sessions of the conference.


tyrasd:
@Wilm0r: does this look ok or is there anything to be changed still?

Wilm0r:
Sorry for the delay! You caught me in a wave of travel, not spending much time at home for a bit :-(

But one issue here: The URL is sadly used as a primary key, which means that your change will effectively result in duplicate entries for those who already reloaded the menu json file with its current revision. Is that a problem? Do the two have helpfully different titles at least to avoid confusion?

(Yes, this is an annoying trait of Giggity, but sadly Pentabarf have nothing like an identifier field that would be suitable instead of abusing the URL for that purpose. One solution would be to use the menu.json file AS the menu instead of just as a seed to populate the internal database. Maybe I should really start doing that..)

tyrasd:
> The URL is sadly used as a primary key, which means that your change will effectively result in duplicate entries for those who already reloaded the menu json file with its current revision. Is that a problem?

Ah, ok. I guess that not too many attendees would have installed the app already, since we haven't yet advertised it publicly (in the conference booklet we do however have a QR-code to giggity and the new schedule xml). IMO it's fine to go ahead and merge this anyway.

Wilm0r:
I see the titles are exactly the same though, still. Giggity right now will show no clear difference between the two, which again is a ~10yo design decision that I should revisit ... but in the meantime, are you able and willing to add "(merged)" to the in-file title (which overrides what's in menu.json) of the new file?

tyrasd:
> I see the titles are exactly the same though, still.

Ah, I see. I would rather change the title of the original non-merged schedule to read "State of the Map 2019 – General Tracks".  We can do this in the pretalx system, but I don't have the necessary admin rights for it. I'll ask the person in charge for it to change it and let you know once that is done.

Wilm0r:
Ack, thanks :-) If that fails, I guess we can try our luck here but it *will* be confusing to those who may have had Giggity installed already from a different conf for example.

tyrasd:
The old schedules have been renamed now.

tyrasd:
I think this can be merged now @Wilm0r?

Wilm0r:
Sorry for the delay, was offline for a while.

tyrasd:
just in time :)

PR #122, opened 2019-08-21
Add CCCamp 2019 full schedule

mmuman:
URL taken from https://events.ccc.de/camp/2019/wiki/Static:Schedule


Wilm0r:
Sorry, was not paying much attention for a while due to travel and busy times at work, though really <1d was not enough notice anyway. :-(

PR #123, opened 2019-08-21
Add LVEE 2019 to the menu

andrewshadura:
This adds [LVEE 2019](https://lvee.org/).

It would be great if this could be merged sooner rather than later since it starts tonight :slightly_smiling_face: 

Thanks!


andrewshadura:
Ping?

andrewshadura:
There’s no need for this.

PR #124, opened 2019-09-05
Added MRMCD 2019

joker234:
Tested with `ci/menu.py` and the qrcode command in the README :)


Wilm0r:
Oops, sorry for not having caught this one in time :-( Was offline for non-tech fun stuff for the past week and a bit. Taking myself out of the loop for merges like this is still on my TODO (the split + merge scripts and CI were one step into this direction).

PR #126, opened 2019-09-05
Switch to python3 for QR code generation

joker234:
Since Python 2 is discontinued very soon (https://pythonclock.org), you might want to switch to Python 3 for all parts of your project. This PR just changes the QR code generation advice in the README.


Wilm0r:
(Seems dumb to have the python version number not only in the pkg but also in the command name but I see there's no symlink other than a /usr/bin/qr alternative (with no guaranteed flags like --optimize so useless))

Thanks!

PR #129, opened 2019-09-18
add Ubucon Europe 2019

tcarreira:
tested with ci/menu-ci.py

Ubucon Europe will happen between 10-13 oct 2019 in Sintra (Portugal).
Would be nice to be listed in the menu.

Thanks

### DIFF
--- a/menu/ubucon_europe_2019.json
+++ b/menu/ubucon_europe_2019.json
@@ -1,0 +1,58 @@
+{
+  "version": 2019092101,
+  "url": "https://manage.ubucon.org/eu2019/schedule/export/schedule.xml",
+  "title": "Ubucon Europe 2019 - Sintra",
+  "start": "2019-10-10",
+  "end": "2019-10-13",
+  "metadata": {
+    "links": [
+      {
+        "url": "https://sintra2019.ubucon.org",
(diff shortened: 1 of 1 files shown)


Wilm0r:
Yes, we can include this :-)

I noticed BTW that you also have an XML formatted schedule which is *much* preferrable: https://manage.ubucon.org/eu2019/schedule/export/schedule.xml

Could you update your PR to use that?

tcarreira:
Hey @Wilm0r, sure. I just updated the commit and force-pushed (so we have only one commit for this new event).
Let me know if you need something :)

Thanks

Wilm0r:
I'm sorry for the annoyance, but could you please also remove the "id" field? It's not needed and actually problematic for XML files. This is because of a >10y old Giggity design flaw that I intend to fix hopefully soon. https://github.com/Wilm0r/giggity/issues/134

Note by the way that force pushing isn't necessary since I can simply use the "Squash and merge" feature here to combine your commits into a single one on my side, so that should be simpler for you.

tcarreira:
Sure, I just removed the id. 
Actually I forgot the github's r&s feature. Well, I'll keep it with the force push now.

Thanks

PR #130, opened 2019-09-19
⌛ Add Datenspuren 2019 to the menu (this weekend)

astro:
I hope this gets seen and merged within the next two days. Otherwise close as irrelevant.

### DIFF
--- a/menu/datenspuren_2019.json
+++ b/menu/datenspuren_2019.json
@@ -1,0 +1,16 @@
+{
+	"version": 2019092000,
+	"url": "https://datenspuren.de/2019/fahrplan/schedule.xml",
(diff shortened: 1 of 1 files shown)


mmuman:
> Icon for Datenspuren 2019 seems bad: 400×480 image is not square

You can either drop the icon or use [your twitter avatar](https://pbs.twimg.com/profile_images/1115200342654894087/KFpFqQBb_200x200.png).

mmuman:
Btw, remember you can also print QR codes with the URL of the schedule for people to scan directly with Giggity on location.

Wilm0r:
I've merged mmuman's version of this. People will need to pull down to refresh but then it'll show up.

PR #132, opened 2019-09-20
Add Open-Access-Tage 2019

tullney:
Adding Open-Access-Tage 2019, the annual German-speaking conference on open access to scholary publications.

See

* http://open-access-tage.de/2019
* https://twitter.com/oatage


tullney:
Thanks for the super-fast merge!

PR #136, opened 2019-10-01
Add Paris Web 2019

mmuman:


mmuman:
Oddly travis says it's fine but github doesn't notice…

menu/parisweb_2019.json
```diff
@@ -0,0 +1,22 @@
+{
+	"version": 2019100200,
+	"url": "https://www.paris-web.fr/2019/conferences-parisweb-2019.ics",
+	"title": "Paris Web 2019",
+	"id": "Conf\u00E9rences Paris Web 2019",
```
Wilm0r: The file should be utf-8 clean BTW, it's 2019 not 1970

menu/parisweb_2019.json
```diff
@@ -0,0 +1,22 @@
...
+	"speakermatch": " \u2014 (.*)",
```
Wilm0r: Huh, that never got merged right? But sure, if CI lets you, then why not ¯\_(ツ)_/¯

mmuman: Oh no, never finished the patch for this, I just copy-pasted from last time.

menu/parisweb_2019.json
mmuman: I think I had an issue with the accented char last time, not sure which it was.

mmuman: Oh, indeed it's json_pp (that I use to check the syntax and pipe the output to gzip|python3-qr) which is buggy, on another menu it transforms "Règlement" quite badly so it appears broken in the menu pane:
> 000000e0  65 22 20 3a 20 22 52 c3  83 c2 a8 67 6c 65 6d 65  |e" : "R....gleme|
> 000000f0  6e 74 22 2c 0a 20 20 20  20 20 20 20 20 20 20 20  |nt",.           |

mmuman: Ah, adding utf8 to -json_opt fixed it. Bleh.

PR #139, opened 2019-10-23
Add Alchimie 13 demoparty

mmuman:
Probably only useful for me but who knows ;-)

### DIFF
--- a/menu/alchimie_13.json
+++ b/menu/alchimie_13.json
@@ -1,0 +1,25 @@
+{
+	"version": 2019102300,
+	"url": "http://triplea.fr/alchimie/documents/alchimie13.ics",
+	"id": "Alchimie 13",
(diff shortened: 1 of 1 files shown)


mmuman:
> Serveur de vote link for Alchimie 13 appears broken: <urllib3.connection.HTTPConnection object at 0x7f6efd935c88>: Failed to establish a new connection: [Errno -2] Name or service not known

Oh dear. This is on purpose: it will be a local machine with a local IP during the event… (of course there is now an official .party TLD for who knows why but still…)

Wilm0r:
> Oh dear. This is on purpose: it will be a local machine with a local IP during the event… (of course there is now an official .party TLD for who knows why but still…)

Because ICANN-- love one-time unsustainable goldrushes turning the root zone in yet another just very expensive TLD. :<

PR #140, opened 2019-10-23
Add PyCon FR 2019

mmuman:
Unable to test due to SSL error on Android 4 though.

### DIFF
--- a/menu/pycon_fr_2019.json
+++ b/menu/pycon_fr_2019.json
@@ -1,0 +1,20 @@
+{
+	"version": 2019102401,
+	"url": "https://www.pycon.fr/2019/pyconfr-2019.ics",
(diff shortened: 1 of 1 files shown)


Wilm0r:
Hah @ Travis error. Guess it's running in a US timezone. I'll look into that, ignore it.

Wilm0r:
Also, out of curiosity, what Android 4 + TLS error is that? Sounds bad..

mmuman:
Unsupported cypher I suppose, and HTTP redirects anyway…

Wilm0r:
> Hah @ Travis error. Guess it's running in a US timezone. I'll look into that, ignore it.

Hmm no it runs in UTC so it's not completely lying. Maybe I'll add some tolerance to it.

PR #142, opened 2019-10-28
Add OSSEU+Linux Embedded Conference Europe

mmuman:
Combined shedule.

Stupid sched.com puts a "X-WR-CALDESC:Event Calendar" in the ICS so the conf gets renamed…

### DIFF
--- a/menu/oss_eu_2019.json
+++ b/menu/oss_eu_2019.json
@@ -1,0 +1,20 @@
+{
+	"version": 2019102800,
+	"url": "https://osseu19.sched.com/all.ics",
+	"title": "Open Source Summit + Embedded Linux Conference Europe 2019",
+	"id": "osseu19",
+	"start": "2019-10-28",
+	"end": "2019-11-01",
+	"metadata": {
+		"links": [
+			{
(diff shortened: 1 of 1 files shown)


mmuman:
Error 500… odd, works here…

Wilm0r:
Yes, that is strange. I can fetch it fine as well with at least wget.
Running the CI script locally also does not reproduce?

mmuman:
Ah, same here… I don't have the reflex to run it everytime. Maybe the User Agent it uses is blacklisted or something?

Wilm0r:
The urllib3 default appears to be no User-Agent: header at all and
adding one indeed seems to fix this. Weird. Guess I'll add that when I
have a chance..

Wilm0r:
> Stupid sched.com puts a "X-WR-CALDESC:Event Calendar" in the ICS so the conf gets renamed…

That's fixed in the quagmire branch BTW, at last. But yes that's not going to help current users. :-( It's not really release-ready yet and it will drop API <6.0 (23?)... Don't really fancy a backporting exercise though. :-/

Wilm0r:
Oh drat, only now notice that the event has already started..

PR #145, opened 2019-12-15
Add 36C3 main rooms

mmuman:
Rooms are checked against last year's c3nav as this year's not public yet.

### DIFF
--- a/menu/36c3_main_rooms_only.json
+++ b/menu/36c3_main_rooms_only.json
@@ -1,0 +1,53 @@
+{
+	"title": "36c3 - main rooms only",
+	"metadata": {
+		"links": [
+			{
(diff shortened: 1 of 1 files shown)


Wilm0r:
Hrmm Giggity should do a better job at menu.json updates now, so maybe we should just merge this now anyway?

Let me click the ready button to check whether CI is happy other than the c3nav 403'ing.

Wilm0r:
WTF for some reason I had to run it by hand, why is Travis such a useless turd? Anyway, erroring on c3nav 403'ing plus apparently the xml not serving when there isn't a User-Agent header, brilliant as well. Merging, thank you! \o/

mmuman:
Thanks!

PR #146, opened 2019-12-25
Fix 36c3 merged URL

jspricke:


Wilm0r:
Urgh, I don't think I can merge this anymore. :< The link works as-is and it's that way on the Wiki, though indeed the https one works.

Problem is: Internally Giggity still uses the URL as the unique identifier of a schedule. So if this gets merged, an additional 36c3-merged schedule will appear in the main menu, which is confusing and annoying. (It could be tagged with "(HTTPS)" but how helpful is that?)

Given the kind of conference this is, doing anything cleartext on wireless there is a sad thing to do obviously. (-: But .... I'm afraid enough people have likely picked up the wrong URL already by now.

I think the best thing I can do right now is add code to the CI to warn when non-HTTPS URLs are added (at least when they have HTTPS equivalents). (Also, not using URLs as the only unique identifier probably would be good yes!)

jspricke:
For me the http version results in: "Network I/O problem: java.io.IOException: Cleartext HTTP traffic to data.c3voc.de not permitted"
I'm not sure what is blocking it (LineageOS 16, no special filter or firewall), but I guess I'm not the only one. So maybe we should add it anyhow?

Wilm0r:
Drat. OK, thanks for spotting and reporting. I'll update it then, and just mark the existing one (which then at least wasn't used by pretty much anyone probably!) appropriately..

Wilm0r:
I wonder what fed you a HSTS header for the domain BTW. If I load the HTTPS equivalent in my browser then go back to HTTP, I don't get redirected.

Anyway, about to push these commits. And I guess I'll need a way to support deletes again/have non-URL primary keys. (I had that, but in a way that didn't work so I've just finished cleaning that up..)

jspricke:
I guess one of the other Fahrplan apps in F-Droid, maybe. Anyway, thanks for fixing it!

Wilm0r:
Possibly, though I saw the same thing on my Pixel 3 on which I don't remember ever installing any of those. Mysterious, but I hope this fix will do and not cause too much confusion.

PR #150, opened 2020-02-10
Add Bibliothekartag 2020 to menu

zuphilip:
Add Bibliothekartag 2020 to menu which uses a Pentabarf XML created by https://github.com/UB-Mannheim/bibtag-scheduler

### DIFF
--- a/menu/bibtag20.json
+++ b/menu/bibtag20.json
@@ -1,0 +1,25 @@
+{
+	"version": 2020021001,
+	"url": "https://cloud.bib.uni-mannheim.de/index.php/s/AwMc2ao8C2fX8rJ/download",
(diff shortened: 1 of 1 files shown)


Wilm0r:
While I have you here :-) Could I convince you to submit an update of the German translations as well?

zuphilip:
Thank you very much for the fast merge!

Certainly, I can look at the German translations, but they already look okay. What I can see are the following points:
- `reminder_enabled_sub_off` was deleted in the English file -> Does this mean we should delete it in the other languages as well?
- The strings `navdrawer_on` and `navdrawer_off` are untranslated yet. But I haven't seen them in the app. In which context do they occur?
 - Anything else?

PR #158, opened 2020-02-28
Allow schedules hosted on http websites

Zverik:
This attribute change allows using schedules that are hosted on non-ssl enabled websites. Currently Android 9+ forbids this without an explicit permission.

Re: [stackoverflow](https://stackoverflow.com/questions/45940861/android-8-cleartext-http-traffic-not-permitted) and [attribute documentation](https://developer.android.com/guide/topics/manifest/application-element#usesCleartextTraffic).


Wilm0r:
O_o Thanks for noticing and the PR.

Do you need this built and released near-term?

Zverik:
No, for the nearest conference I've moved my schedules to an https-enabled location, so there's no hurry. But it'd be useful for future conferences.

Zverik:
Thank you for merging!

PR #161, opened 2020-03-10
Add FOSSGIS 2020

simon04:


Wilm0r:
The version# was wrong. Also, the conf is over now it seems.

PR #163, opened 2020-07-16
WIP: Add HOPE 2020

mmuman:
HOLD: URL is temporary, final URL should be available at the end of the week.

Tested, works even on my Android 4 tablet, for once.


mmuman:
They published the schedule online the other day, and the URL we use is the one pretalx publishes in the export menu. So I think we can go forward and use it anyway, as I didn't hear from them yet, I guess they are too busy doing preparation.

PR #164, opened 2020-08-09
Add Debconf20

stefanor:
Add DebConf20, retire DebConf19.


Wilm0r:
Drat, this one's going to suck with Giggity still not having timezone awareness, isn't it..

stefanor:
> Drat, this one's going to suck with Giggity still not having timezone awareness, isn't it..

Yeah, probably.

Looking at our XML, we've also some weird "day" values, which at least Giggity doesn't seem to be falling over.

Wilm0r:
OK, sorry for the delay. Let's see whether I can get very basic tz awareness added in time. It's going to be next to useless without that. Merging this now already though.

It'd be nice if schedules had a timezone name in them that I could show next to the date or something. But I guess I should be happy they've picked up tz awareness at all (Pentabarf did not have that info back when I first wrote the parser, I'm reasonably sure).

Wilm0r:
(BTW 2.0.6 should fix the tz issues.)

PR #165, opened 2020-08-18
Add FrOSCon 2020

mmuman:
Just noticed FrOSCon is this week-end :)

Removed the map links, hopefully we can add them back next year.


Wilm0r:
> Removed the map links, hopefully we can add them back next year.

Oh yes please. I, too, look forward desperately to normality.

Conveniently thanks to the Debconf20 PR I've added tz awareness to Giggity over the weekend, hope to push that out before the weekend. as well. \o/

PR #171, opened 2020-10-13
SFScon 2020

madeincima:
Added SFScon 2020 ( South Tyrol Free Software Conference ) json in menu


Wilm0r:
Retarded Travis-CI stopped showing the results here directly, but *is* running the test, showing:

```
jsonschema.exceptions.ValidationError: 20201113 is less than the minimum of 2008010100
```

Could you please format the version correctly? (And I hope CI integration will work again..)

madeincima:
Hello @Wilm0r ! After many commits, I think (hope) that everything's all right :-) Thanks for your support!

Wilm0r:
Excellent, thanks :) (And glad the CI is working again.. BTW you can just run `tools/menu-ci.py` before committing to save yourself some time, or did that not work?)

I see this is an iCal file. Note that Giggity's timezone awareness (which as you can imagine became a very interesting feature this year..) currently only works properly with Pentabarf files, but I'll see whether I can improve that, looks like I've got about a month.

PR #174, opened 2021-01-13
Add FOSDEM 2021

mmuman:
Much simpler than usual since we won't need to locate rooms physically
this year.

### DIFF
--- a/menu/fosdem_2021.json
+++ b/menu/fosdem_2021.json
@@ -1,0 +1,25 @@
+{
+	"version": 2021011300,
+	"url": "https://fosdem.org/2021/schedule/xml",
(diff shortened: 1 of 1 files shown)


Wilm0r:
Woo, thanks! Just noticed this morning that the (draft) XML schedule is ready, yes. Easy to forget it's getting this close when I don't need to book train tickets for it....

Why WiP BTW? I do guess there may be some more useful links to add in soon. (And I wonder whether I should add more native handling of streaming links with specific events, hm.)

mmuman:
Yes, maybe we'll get a list of video streams.

mmuman:
I guess we can always update it anyway, let's merge this.

PR #179, opened 2021-06-29
Add Paris Web 2021

mmuman:
They're back!


Wilm0r:
Yay! A non-virtual conference? Is this real? :D

PR #180, opened 2021-07-01
Add State of the Map 2021

joker234:
Sorry that we're a little bit late this year, but it would be really great if this could be merged soon :)

### DIFF
--- a/menu/sotm_2021.json
+++ b/menu/sotm_2021.json
@@ -11,14 +11,6 @@
                                 "url": "https://2021.stateofthemap.org/",
                                 "title": "Website"
                         }
-                ],
-                "rooms": [
-                        {
-                                "name": "Track 1 - Talks"
-                        },
-                        {
-                                "name": "Track 2 - Panels and Workshops"
-                        }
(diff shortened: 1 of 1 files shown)


menu/sotm_2021.json
```diff
@@ -0,0 +1,24 @@
...
+        "end": "2021-07-11",
+        "metadata": {
+                "icon": "https://2021.stateofthemap.org/img/sotm_2021-logo-small.png",
+                "links": [
+                        {
+                                "url": "https://2021.stateofthemap.org/",
+                                "title": "Website"
+                        }
+                ],
+                "rooms": [
```
Wilm0r, line 15: This is not necessary by the way if you're not specifying any metadata. It's all in the schedule file. Listing rooms in the JSON is useful if you want to specify locations for them, etc. Want to remove it again? (I should add this to the test..)

joker234: Thanks for the suggestion, I used sotm-2019 as a template, when we had rooms with actual locations :see_no_evil:

PR #181, opened 2021-07-18
Add Entrée Libre 2

mmuman:
French FLOSS event.


mmuman:
Dang it, Cozy Cloud puts unique temporary IDs in the direct download link, so it won't work :-(

mmuman:
Ok, should be fine now.

Wilm0r:
Sorry for the delay, I see I'm not late yet fortunately!

PR #189, opened 2021-09-17
[Best-Before: 2021-09-20] Add Datenspuren 2021

astro:
Just in time. As always. Thank you.


Wilm0r:
Grr, tedious that the CI didn't run automatically yet... did you get notification about the failure?

The `timzeone` property is new and not actually used by any released version of Giggity yet, but care to add it anyway?

PR #190, opened 2021-09-22
Add SFScon 2021

LucaMiotto:
Add information about South Tyrol Free Software Conference 2021


Wilm0r:
Thank you! And gh Actions NOT running automatically is quite a step back from Travis CI (except that the latter stopped working entirely of course), thanks for the 302 fix!

PR #191, opened 2021-09-27
add MatomoCamp

Findus23:
Thanks for the great app that already helped me in the past with large events.

I hope the version number is correct, I wasn't quite sure how to create it.


Wilm0r:
Yep, version# is correct! The CI hopefully gave you a good suggestion already.

(Also I think I could remove that field by now since it's quite tedious and no longer needed by the new database code.)

Anyway, merging, thank you/you're welcome!

PR #192, opened 2021-12-27
Add rC3: NOWHERE 2021

mmuman:
Just for the record as the event is already started.


Wilm0r:
(Let's merge to keep it in history anyway..)

PR #193, opened 2022-01-06
Add FOSDEM 2022

mmuman:


Wilm0r:
Thank you as always and sorry for the delay!

PR #197, opened 2022-03-02
Add JRES 2022

mmuman:
We don't need the `id` field anymore for ICS, right?


Wilm0r:
Correct, the stupid `id` field is finally gone!

PR #198, opened 2022-03-06
Check for null timezone when loading schedule.

michaelgrahamevans:
When refreshing a schedule for a custom URL, `tz` is null and and a `NullPointerException` is thrown.


Wilm0r:
Yikes, sorry about that one! To avoid crashes while this is not yet released, you could BTW enter your schedule as a JSON fragment with metadata incl tz and URL pointing at whatever you already had.

PR #199, opened 2022-03-08
JRES 2022: Add icon

mmuman:
The image is on a Mastodon instance, should be stable enough.


Wilm0r:
Icon for JRES 2022 seems bad: 323×322 image is not square

Gosh I guess that's square enough anyway.. Let's just merge?

mmuman:
I've asked them if they can make another one…

Wilm0r:
Nah, no need, conf organisers are probably busy enough already!

PR #202, opened 2022-04-04
Add FOSSASIA Summit 2022

mmuman:


mmuman:
Ah, icon is JPEG…

mmuman:
Dropped the icon.

Wilm0r:
Hrm, yeah, so a while ago @heysadboy contributed a parser for FOSSASIA's JSON format which seemed like it was trying to become a broader standard. IIRC I've removed that code again a while ago since I couldn't see even FOSSASIA themselves still using it. :(

Sad because .ics stinks. But so be it.

And yeah, I think .jpeg icons *would* work but let's just not do it. Lack of transparency--

mmuman:
Yeah, shame but well…

PR #204, opened 2022-05-01
Add Bibliothekskongress 2022

zuphilip:
I am having some troubles to add the json file from another location directly, but hope this will work fine from the menu here. The pentabarf file works fine. However, updating a schedule where the pentabarf file is added directly seems not yet to work in the current version, because the fix https://github.com/Wilm0r/giggity/commit/9c404210fa9524c7caab3a209fb21a3c26b21311 is not yet part of the released version.


zuphilip:
friendly ping @Wilm0r

zuphilip:
Thank you very much!

Wilm0r:
np and thank you for contributing!!

PR #208, opened 2022-05-28
Add Electromagnetic Fields 2022 (EMF2022) to the Giggity list

leyrer:
Electromagnetic Fields is happening. See https://www.emfcamp.org/


Wilm0r:
Wow, EMF is happening already. I had 2020 tickets ... but not in the area now :( Have fun!

PR #211, opened 2022-06-07
Add timezone to JSON menu example in README

EdwardBetts:


Wilm0r:
Ha good idea, thank you! But it sits in between `refresh_interval` and its comment, can you fix that?

EdwardBetts:
Done.

PR #213, opened 2022-06-08
Add Pass The Salt 2022

mmuman:


mmuman:
FTR, it starts today.

Wilm0r:
O_o I thought I flushed the PR queue quite recently. Sorry once again.. :(

(I've looked into giving you commit access in the menu subdir, there seems to be no way though. Could just do full repo instead..)

PR #214, opened 2022-06-14
Add DebConf 22

EdwardBetts:

PR #216, opened 2022-07-05
Add DebConf 22 icon

fortysixandtwo:
See https://github.com/Wilm0r/giggity/pull/214


fortysixandtwo:
@Wilm0r would adding an alpha channel suffice (because it still complains about the size)?

PR #220, opened 2022-08-09
Update source of State of the Map 2022 schedule: adds missing academic talks

tyrasd:
The schedule on https://pretalx.com/sotm2022/ is missing the academic track talks (on Sunday in Auditorium B, see https://2022.stateofthemap.org/programme/#general_Sunday), because the academic talks are managed in a separate pretalx instance (https://pretalx.com/state-of-the-map-2022-academic-track/ vs. https://pretalx.com/sotm2022/). For this reason, we maintain a merged schedule xml during the conference days available at https://sotm.osmz.ru/2022.xml

### DIFF
--- a/menu/sotm_2022.json
+++ b/menu/sotm_2022.json
@@ -1,5 +1,5 @@
 {
-        "version": 2022060701,
+        "version": 2022080901,
         "url": "https://sotm.osmz.ru/2022.xml",
         "title": "State of the Map 2022",
         "start": "2022-08-19",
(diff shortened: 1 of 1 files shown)


tyrasd:
//cc @MarcoMinghini @Zverik

Zverik (approved):

Wilm0r:
Uhoh, so the URL is used by Giggity as a primary key for selections of talks to attend for example... I used to have a convoluted separate unique ID mechanism for this, but this caused more trouble than it was worth.

So if I merge this, anyone who already picked talks to attend will lose that. So maybe we should instead have a separate merged schedule so peple can pick?

Zverik:
We haven't yet announced the schedule in your app, so there'll be a most couple people affected, which we can afford.

Wilm0r:
OK that's perfect :)

PR #223, opened 2022-08-23
Add Capitole du Libre 2022

mmuman:
It's back \o/

Schedule isn't available yet.


mmuman:
Schedule is published, just tested working.

Wilm0r:
Pfew, still caught this in time!

PR #224, opened 2022-08-24
Add OAT22 to the menu

zuphilip:
I created a skeleton of the schedule (manually) as an XML file and updated it with the abstracts from the website. The icon is from twitter, but I needed to add an alpha layer (hope this works now).

### DIFF
--- a/menu/oat22.json
+++ b/menu/oat22.json
@@ -1,0 +1,17 @@
+	{
+		"version": 2022082401,
+		"url": "https://cloud.bib.uni-mannheim.de/s/f35r5JMRZywZBoa/download/oat22.xml",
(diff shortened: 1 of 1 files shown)


Wilm0r:
Thank you!

I think the icon may look a little tiny and it'd probably be better to crop it a little. If you do so, note that Giggity does pretty naive TTL-ignorant caching so then we'll have to update the JSON.

Not super important anyway, icon's only used in the Android activity scroller and if the user creates a home shortcut.

zuphilip:
Thank  you very much for merging. The icon looks reasonably good for me, but I will try to remember your advice for the next time.

PR #228, opened 2022-09-09
Add Pi and Radio / UKW-Tagung 2022

danielfett:
This PR adds our event Pi and Radio / UKW-Tagung 2022 to the menu.

Since we are late with submitting this - sorry! - a quick merge would be highly appreciated!


Wilm0r:
Sorry for missing this :(

Note by the way that you can get all of this without actual menu integration using `tools/ggt.sh` as explained on https://github.com/Wilm0r/giggity/ . Just put that deeplink on your page + maybe same link in a QR so people can scan it with their phone if they're on your website with a less mobile device.

PR #232, opened 2022-09-27
Added new entry for PyConES 2022

pablomartinez:
Added a new entry for the PyCon conference in Spain


Wilm0r:
Bit late but maybe this is still useful :(

PR #233, opened 2022-10-01
Add Paris Web 2022

mmuman:


mmuman:
Let's drop the icon for now.

Wilm0r:
IIRC it lacked an alpha layer? Not *that* big a deal realistically BTW. We could see how it looks some other time.

PR #243, opened 2023-01-29
Create hackerhotel_2023

qwaxys:


qwaxys:
@Wilm0r it should validate but the year wasn't updated to 2023 yet, so I fixed that too.

Wilm0r:
Hahaha I forgot I ever even did that! https://github.com/Wilm0r/giggity/blob/master/tools/menu-schema.json#L17

Thank you for finding that. I'll set it to 2030 or so myself. The CI already enforces this more properly (to current date instead of a constant).

And I think some day I should be able to just drop the silly serial#, or at least not bother submitters with it..

Wilm0r:
Fixed with a few more years headroom than just 1 :) https://github.com/Wilm0r/giggity/commit/32cc3697ea8204c13603b34a5a53337d6b6ce4eb

I notice your file lacks an extension, and apparently I already failed to notice that on one other submission, oops ... but would you mind to correct this? I'll add validation for this too.

PR #245, opened 2023-01-30
add CfgMgmtCamp 2023 Ghent edition

toshywoshy:
This adds CfgMgmtCamp 2023 Ghent Belgium edition

Signed-off-by: Toshaan Bharvani <toshaan@vantosh.com>

### DIFF
--- a/menu/cfgmgmtcamp_2023_ghent.json
+++ b/menu/cfgmgmtcamp_2023_ghent.json
@@ -5,7 +5,6 @@
 	"start": "2023-02-06",
 	"end": "2023-02-08",
 	"timezone": "Europe/Brussels",
-	"refresh_interval": 1800,
 	"metadata": {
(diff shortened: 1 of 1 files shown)


Wilm0r:
BTW background on this error:

Schedule CfgMgmtCamp 2023 Ghent BE refresh_interval set below 1d while server never sends HTTP 304s

That one means that you're trying to set a pretty rapid refresh time while your web server isn't configured properly to send a HTTP 304 if there are no changes to the file. It's an optimisation I've made to deal with often faulty conference WiFi, though perhaps that's becoming a dated assumption.

toshywoshy:
I removed the refresh so we can push before the event starts

Wilm0r:
Sorry for not getting back in time, but also the CI was *still* failing, is Github's warning about that not clear enough? :-/

PR #256, opened 2023-03-16
Add FOSSASIA Summit 2023

mmuman:


mmuman:
Not sure what's up with the CI…

Wilm0r:
Yeah no worries that's a change I made that is having dumb Github issues. :(

Wilm0r:
It's unfortunate BTW that FOSSASIA don't seem to have a JSON schedule anymore, Giggity had support for it but I've just ripped that out anyway. If you truncate the schedule URL from .json on then you do get some JSON data but not enough detail. :(

PR #258, opened 2023-03-25
Add JdLL 2023

sparunakian:
Hi! This MR adds the 2023 edition of the JdLL event.


sparunakian:
It seems that the failure of the second batch of checks is not due do this MR.

Wilm0r:
Yep sorry, really need to fix that :( Trashy github permissions model..

mmuman:
Oh cool, I was just noticing I didn't do it this year, then I saw this when pulling. Thanks!

sparunakian:
Thank you for merge and thank you for this project!

PR #262, opened 2023-04-05
Added Easterhegg 20

qwaxys:


Wilm0r:
(Yay @ the bot just working \o/ )

Wilm0r:
Hehe, thanks :) Thinking of adding a "conference starts in x days" message to it so that I'll immediately know which ones are more urgent.

qwaxys:
Happy to help if you need more maintainers :)

You should be able to know if a conference is soon since there's beginning and end field?

PR #264, opened 2023-04-09
Add THSF 2023

mmuman:


mmuman:
`cannot identify image file` odd, it's a plain PNG file…

Wilm0r:
Huh, oh fun, and the posted comment is not exactly helpful either in this case. Does the CI checker work if you run it locally?

Wilm0r:
Does look like the file is corrupted in ways that some readers will ignore and others will not...

```
>>> PIL.Image.open(open("favicon.png", "rb"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/PIL/Image.py", line 3283, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file <_io.BufferedReader name='favicon.png'>

...

wilmer@ruby:/tmp$ convert favicon.png favicon2.png 
convert-im6.q16: iCCP: CRC error `favicon.png' @ warning/png.c/MagickPNGWarningHandler/1668.
wilmer@ruby:/tmp$ ls -la favicon*
-rw-r--r-- 1 wilmer wilmer 32K  9 apr 17:57 favicon.png
-rw-r--r-- 1 wilmer wilmer 30K  9 apr 19:11 favicon2.png

...

>>> PIL.Image.open(open("favicon2.png", "rb"))
<PIL.PngImagePlugin.PngImageFile image mode=RGBA size=100x100 at 0x7FD6ABDB0990>
```

A 100×100 PNG being >30KB is pretty odd anyway.. But huh also look at this:

```
wilmer@ruby:/tmp$ exiv2 favicon.png 
File name       : favicon.png
File size       : 31771 Bytes
MIME type       : image/png
Image size      : 100 x 100
Thumbnail       : image/jpeg, 17096 Bytes
```

So more than half of it is a thumbnail in JPEG?

Meh it all doesn't really matter that much if you can confirm that the Android PNG handler takes the file properly?

mmuman:
I'll see if they can fix it.

Wilm0r:
If Android reads the file without complaining let's not bother them BTW :)

mmuman:
Still testing on Android 4 so I don't really know.
Plus it's probably not the correct one anyway (it's the logo of the organizers, not the event).

mmuman:
Removed the icon, we can add it back if needed.

mmuman:
I guess CI doesn't like having both PSES at same URL?

Wilm0r:
Huh hmm. I guess the CI actually uses URL as a primary key too. ... yep! https://github.com/Wilm0r/giggity/blob/master/tools/menu-ci.py#L279

But OK it's a CI not a cop. Let's merge.

PR #267, opened 2023-04-16
Add PSES 2023

mmuman:


mmuman:
Hmm yeah it's probably not a good idea to have the same URL for each year… maybe with a fragment to fake a unique URL?

mmuman:
Ok, forgot to rename from 2019 :-D

mmuman:
Dropped the live link, we can always add it back.

Wilm0r:
> Hmm yeah it's probably not a good idea to have the same URL for each year… maybe with a fragment to fake a unique URL?

Yikes. I don't think I've written a check for that actually? But yeah that sucks, since Giggity still uses those URLs as a primary key. :( But I guess talks will have unique ID's so no ghost selections are likely. And heck I'd be impressed if enough people even have the same phone like then (and with Giggity still installed all this time).

PR #271, opened 2023-04-26
Link to xml file of frab export

michael-hartmann:
It seems that the link should go to the xml version, not the json version, This PR fixes the mistake.


Wilm0r:
Ha, oops, yes this is correct. Giggity doesn't know that JSON Format yet. Is that Frab/pretalx/etc JSON, hopefully roughtly the same just with fewer angle brackets?

michael-hartmann:
I think the xml and json are essentially the same, but never really checked. I am mainly using Json because wirh xml I usually don’t see the forest for the trees :-)

PR #274, opened 2023-05-15
Add Opensouthcode 2023

davidsf:
Add Opensouthcode 2023, June 9 and 10 in Málaga (Spain)


Wilm0r:
Thanks for the addition :) And hm, maybe I should make the redirect check less strict..

PR #283, opened 2023-07-05
FOSSY 2023: Use schedule in XML converted from the JSON version

EdwardBetts:
Reads a version of the schedule in XML that has been generated from the JSON version of the schedule.

The XML version is being regenerated regularly.

This gives more detail that the ICS version.

### DIFF
--- a/menu/fossy_2023.json
+++ b/menu/fossy_2023.json
@@ -1,7 +1,7 @@
 {
 	"version": 2023070501,
 	"url": "https://edwardbetts.com/fossy_2023/schedule.xml",
-	"title": "FOSSY 2023",
+	"title": "FOSSY 2023 (use this one!)",
 	"start": "2023-07-13",
 	"end": "2023-07-16",
 	"timezone": "America/Los_Angeles",
--- a/menu/fossy_2023.json
+++ b/menu/fossy_2023.json
@@ -1,6 +1,6 @@
 {
-	"version": 2023070401,
-	"url": "https://2023.fossy.us/schedule/conference.ics",
+	"version": 2023070501,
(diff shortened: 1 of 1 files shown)


Wilm0r:
Oh nice!! On my phone right now, can merge shortly, but one warning: this will invalidate any selections people may have made. Is that still safe?

EdwardBetts:
I think it is safe, the ICS version hasn't been there for long.

Wilm0r:
OK, let's take the risk.

But one more issue, and this is kinda a Giggity bug in a way, but people who used Giggity over the past days will still have the old entries in their menu too. :(

We could either have extra files for the old two URLs with "(old, do not use)" titles which stinks because it'll show up even for people who so far haven't opened Giggity, or maybe just add "(use this one!)" or something to the title of this entry. I think I like that last option more.

I used to use an explicit primary key for schedules instead of their URL but it was tedious in other ways and didn't seem to be worth keeping. I should perhaps reconsider.. IIRC that was before I split the menu entries file with the centralised auto-generator that could now generate these primary keys when absent..

EdwardBetts:
I saw this bug with the switch to the ICS schedule. I agree the second option is better.

Maybe you could use the menu filename (fossy_2023.json) as the primary key.

Wilm0r:
> Maybe you could use the menu filename (fossy_2023.json) as the primary key.

Yeah exactly. That's the best option now with the split menu files. I just need to figure out how to transition this cleanly with bw compatibility etc.

Can you update your PR with option 2 then? Ready to merge whenever you are :)

EdwardBetts:
I've made the change.

PR #289, opened 2023-08-01
Add DebConf 23 in Kochi, India

EdwardBetts:


Wilm0r:
Looks like there's still an issue. Would it help if I have the bot delete-repost instead of editing the existing post, so that it triggers a new email notification?

EdwardBetts:
Thanks, I think probably would've spotted the failing test. I don't think delete-repost is needed.

Wilm0r:
BTW let's see if I can release a new version in time for this fix: https://github.com/Wilm0r/giggity/commit/dd702360abf308bb8ae7ce64878388ba191e7ac2

Bit awkward otherwise.

PR #293, opened 2023-08-11
Add EuroBSDCon 2023

mmuman:
Schedule is for now the one I generated myself from the website. Hopefully they'll host it themselves.


Wilm0r:
Given the conversation in #294 do you have a XML file for this one now?

Wilm0r:
(WTF, justs found this draft comment that actually predates my last one, no idea how that survived in browser/cookie state over days...)

Ha, so I thought I was checking for that, https://github.com/Wilm0r/giggity/blob/9d7869351ec0ff9777be16f34b8038c374d170ef/tools/menu-ci.py#L260 , but I think I failed at my own weird API there since in case of failure it won't return None or whatever, but an exception. (Yes, returns, not throws. Meh.)

.... Anyway, about to fix that now.

Wilm0r:
Didn't you say you found an OSEM XML for this now?

mmuman:
No they said they used OSEM but the schedule doesn't correspond anyway.

mmuman:
I updated and hosted the ics on a Nextcloud server, so we'll get HTTPS.

I added the rooms but it seems they don't work when using ICS, Paris Web also doesn't have clickable room names…

mmuman:
ping, it starts today…

mmuman:
Thanks!

Wilm0r:
Thank you too, as always :D

PR #294, opened 2023-08-11
Add a CSV to ICS helper tool

mmuman:
I wrote this to extract the schedule from the website of the EuroBSDCon: https://2023.eurobsdcon.org/program/

Copy-paste the tables into a spreadsheet (like LibreOffice Calc), export as CSV (, as delimiter, " as quote).

It expects:
- lines that start with the event type and day
- lines starting with an empty cell that lists the tracks,
- lines that start with start_time (eventually end), then each talk.

Lot of hardcoding and debug output to cleanup.


Wilm0r:
Ah yes, looks like I've accidentally deleted it: 76c8dbfdf4600069f63ff4c7fa286e084daeaeba Reason: Primary purpose of that script was once to generate Giggity's original native but long deprecated XML format. I forgot I added a little `frab` method right in the middle of it.

Would you be interested in merging these?

mmuman:
Seems I should have slept instead, they used OSEM for the CfP :sweat_smile: 
Should probably be merged at some point yes.
I went for ICS because I've been working on it these days and I didn't fancy digging XML, plus I didn't have all the required info…

Wilm0r:
What do you want to do with this PR for now?

PR #296, opened 2023-08-11
README: Add hints at locating the XML file from conference sites

mmuman:


Wilm0r:
I'm wondering BTW whether this could be somewhat automated... looked at sitemap files but the schedule info isn't exposed there at least, which is a shame..

PR #304, opened 2023-08-21
Add hack.lu 2023

mmuman:
Schedule isn't published yet.


Wilm0r:
Seems to be there now though?

mmuman:
Indeed

mmuman:
I added the rooms with the coords of the hotel, they might want to fix them next year.

PR #313, opened 2023-09-26
add SFSCON23 event

ila131:
Add SFSCON23, a significant event that will take place on November 10th and 11th, 2023, in South Tyrol.


Wilm0r:
Thank you for adding!

PR #320, opened 2023-10-23
Add Capitole du Libre 2023

mmuman:


mmuman:
Me, before doing the PR: … oh, must fix the timestamp in that JSON file, it's from 2 days ago…

CI: Oh, the timestamp is in the future…

🤷

Wilm0r:
Hah! I worried that CI could be running in some US timezone, but docs day it's UTC, and indeed your comment was on Monday the 23rd at 23:45 UTC I think?

mmuman:
Probably, we're UTC+2 here…

Wilm0r:
I guess I could add a timezone tag in the error message, or for "comfort", maybe use Australian timezone :)

mmuman:
It's fine as-is really.

PR #322, opened 2023-11-01
Add SeaGL 2023

mmuman:
I noticed they use OSEM which does provide the URL and links to Giggity explicitly, let's add it officially :-)

### DIFF
--- a/menu/seagl_2023.json
+++ b/menu/seagl_2023.json
@@ -1,0 +1,22 @@
+{
+	"version": 2023110100,
+	"url": "https://osem.seagl.org/conferences/seagl2023/schedule.xml",
+	"title": "SeaGL 2023",
+	"start": "2023-11-03",
+	"end": "2023-11-04",
+	"timezone": "US/Pacific",
+	"metadata": {
+		"links": [
+			{
+				"url": "https://seagl.org/",
+				"title": "Website"
+			}
+		],
+		"rooms": [
+			{
+				"name": "Room .*",
+				"latlon": [47.65538, -122.30547]
+			}
+		]
+	}
+}


Wilm0r:
> I noticed they use OSEM which does provide the URL and links to Giggity explicitly

Huh wait what do you mean?

menu/seagl_2023.json
```diff
@@ -0,0 +1,22 @@
...
+	"url": "https://osem.seagl.org/conferences/seagl2023/schedule.xml",
+	"title": "SeaGL 2023",
+	"start": "2023-11-03",
+	"end": "2023-11-04",
+	"timezone": "US/Pacific",
+	"metadata": {
+		"links": [
+			{
+				"url": "https://seagl.org/",
+				"title": "Site web"
```
Wilm0r, line 12: That sounds a bit more French than what I'm used to in Seattle? (-;

mmuman:
Either OSEM puts a page on its own or they did, pointing to Giggity on F-Droid with the schedule URL: https://osem.seagl.org/conferences/seagl2023/schedule/app
Still it's less convenient than having it on the menu.

menu/seagl_2023.json
mmuman: oh lol yeah I copied the Capitole file and missed that :-)

mmuman: Fixed

Wilm0r:
Ahh like that :) I am still waiting for some event to use `ggt.gaa.st/#` URLs with the URL *and* JSON fragment in a QR. At which point they don't even really need to submit the JSON here anymore.

(Though probably it's still better to keep doing so too, yes!)

PR #323, opened 2023-11-06
Add BSides London 2023

EdwardBetts:


Wilm0r:
BTW while writing the automated UI tests I noticed that too many metadata links result in scrolling on the nav drawer, which didn't really seem like fantastic UI. How is that looking on your phone with this entry? What do you think about it? I wonder whether to start setting a limit..

EdwardBetts:
I'd spotted the scrolling, it didn't bother me. I tend to add a lot of links, but I don't know if anybody finds it useful or clicks on them.

You make a good point about the scrolling, I'll be more selective with what I include from now on.

PR #326, opened 2023-11-09
Update sfscon_2023.json

LucaMiotto:
Fix remove Attendee and version

### DIFF
--- a/menu/sfscon_2023.json
+++ b/menu/sfscon_2023.json
@@ -1,5 +1,5 @@
 {
-	"version": 2023110898,
+	"version": 2023110899,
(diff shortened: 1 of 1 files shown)


Wilm0r:
Hmm, the issue is that it's not a pdf download but an in-browser viewer? With no direct PDF link even available? O_o

You _could_ try just removing the `type` field by the way, assuming the PDF viewer is mobile-friendly.

Wilm0r:
(https://github.com/Wilm0r/giggity/commit/c225c6b84cabed63592d60dba30b7799c844e999 seems to have worked)

PR #338, opened 2024-03-31
Add JdLL 2024

mmuman:
Seems previous years used Brussel's timezone for some reason… Luckily it's the same as Paris :-D


sparunakian:
The schedule is now public for the 2024 edition: https://pretalx.jdll.org/jdll2024/schedule.xml.

PR #346, opened 2024-04-27
Bump Github Actions

Jean-BaptisteC:


Wilm0r:
Hmm any particular reason to do this now BTW? (Guess just being up-to-date again is a good enough reason of course)

Jean-BaptisteC:
Yes :)
Github drops support to Node JS 16 in Github actions -> https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

Wilm0r:
Huh, my first instinct is "ew node I'm not using that" :P But I presume what this means is that some of the actions/plugins I rely on would stop working?

PR #351, opened 2024-05-17
added IGERla schedule

wuschLOR:
schedule.xml not jet published


wuschLOR:
Event is now in the past

Wilm0r:
Urgh, apologies, your update came right before a long trip for me, and I don't think the update triggered any emails so I never saw it to begin with :(

PR #352, opened 2024-05-17
added Spatial Humanities 2024 schedule

wuschLOR:
schedule.xml not jet published


wuschLOR:
Schedule is now online and ready to be merged :) Many thanks

Wilm0r:
Thanks for the update and for the explicit comment so that it hit my mailbox!

PR #355, opened 2024-05-27
Add GPN22 (Gulaschprogrammiernacht 2024)

asdil12:
Menu Entry for GPN22


Wilm0r:
Whoops, was identical apart from the seq# to https://github.com/Wilm0r/giggity/pull/357 :) But thanks for the contribution too!

PR #359, opened 2024-06-19
Better FOSS4G Europe 2024 schedule

Zverik:
The current version has only the main event schedule and no speaker names. The new one, sourced [from here](https://sotm.osmz.ru/tartu2024.html), merges main event, workshops, birds of feather, all side events, and OGC Innovation Days into one.

I have also added locations for all venues.

### DIFF
--- a/menu/foss4g_europe_2024.json
+++ b/menu/foss4g_europe_2024.json
@@ -32,7 +32,7 @@
 			},
 			{
 				"name": "Room .*|Destination.*|GEOCAT.*|LAStools.*|QField.*|Vanemuise 46",
-        "show_name": "Vanemuise 46",
+				"show_name": "Vanemuise 46",
 				"latlon": [58.37347, 26.71616]
 			},
 			{
--- a/menu/foss4g_europe_2024.json
+++ b/menu/foss4g_europe_2024.json
@@ -32,6 +32,7 @@
 			},
 			{
 				"name": "Room .*|Destination.*|GEOCAT.*|LAStools.*|QField.*|Vanemuise 46",
+        "show_name": "Vanemuise 46",
 				"latlon": [58.37347, 26.71616]
 			},
 			{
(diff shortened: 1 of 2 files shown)


Wilm0r:
Ooh nice, thanks for the contribution :D Did you scrape the website to generate this version yourself?

One issue is that replacing schedules in Giggity is a little messy, especially it's hard to preserve selections. I'm trying to improve this but that work isn't done yet.

Just in case anyone has talks selected already they should be able to find those back. Maybe it's best to mark your schedule with a " (extended)" title suffix?

(Once my work is done we'd be able to do better here, though even then of course only if the extended schedule does use the exact same id's/uuid's for talks, etc.)

Zverik:
I've just used my [schedule merging tool](https://github.com/zverik/schedule-convert), just like for [sotm_2019](https://github.com/Wilm0r/giggity/blob/master/menu/sotm_2019.json) ,[sotm_2021](https://github.com/Wilm0r/giggity/blob/master/menu/sotm_2021.json), [sotm_2022](https://github.com/Wilm0r/giggity/blob/master/menu/sotm_2022.json), [sotm_eu_2023](https://github.com/Wilm0r/giggity/blob/master/menu/sotm_eu_2023.json).

The base component is of course the same xml (actually json, because xml misses speakers), so all the ids, codes, slugs are preserved.

The conference is in two weeks, and Giggity app was promoted second to the official conference app (you know the one, EventFahrplan or something, severely lacking), so I don't think anybody would be affected by the change.

Though if you want I can makr it with "(extended)", but that may bring more confusion.

Wilm0r:
The problem is that if you don't add a tag, people *will* have two *identically* named entries for this conference in their menu which is pretty confusing. So we really need a tag, even though that's indeed also ugly. That's why I'm hoping to fix that behaviour in some future Giggity version.

Wilm0r:
Maybe the best thing to do is actually to keep the existing file as-is but with a " (short/old schedule)" tag.

There's just no great way to deal with this when Giggity uses the URL as the primary key for the event and won't delete anything from its internal database just because the URL changed/disappeared from the menu seed file.

Zverik:
Could you please merge it soon? I'd like to promote the extended schedule in the app among the participants :)

Wilm0r:
Oof, thanks for the ping! Not sure how I missed your last update, either bad luck or that annoying fastmail bug again where it sometimes wrongly marks emails as read. :|

PR #363, opened 2024-07-17
Add basic info about DebConf 24

fortysixandtwo:


fortysixandtwo:
/cc @stefanor and  @EdwardBetts  as you both touched the dc23.json last year (and to avoid duplicated work)

ci check fails because the [DC24 FAQ link](https://wiki.debian.org/DebConf/24/Faq) doesn't exist yet

EdwardBetts:
You can remove the non-existent URL from `menu/debconf24.json` and the CI should succeed.

Let me know if you need any help.

fortysixandtwo:
v2: removed 404'ing link and reduced number of links to 4 to conform to 3e7c21f5506b842a7cd8a029b1df625582ad88e2

stefanor:
LGTM

Wilm0r:
Thanks for shrinking the links section yeah. I guess I could try to auto-truncate to avoid the scrolling (which really is terrible UX I'd say) but 4 probably should suffice in most cases.

PR #366, opened 2024-07-30
add basic hackmas castle 24 event

ruru4143:


ruru4143:
actions fails but doesn't provide an error. 
I checked it via https://jsonlint.com

Wilm0r:
Weird that the comment is empty yes... :-/ but see https://github.com/Wilm0r/giggity/actions/runs/10164485309/job/28109952792?pr=366 for the actual error

ruru4143:
for me, this is ready to merge @Wilm0r

ruru4143:
i will do another pull request as soon c3nav is working.

Wilm0r:
Thank you! BTW I see only 3 or 4 items in the schedule, shouldn't it be a lot more?

Wilm0r:
Wow OK, that's very last minute :)

Sorry, should've asked this earlier, and I'll pay attention to this PR today so that we can get it submitted, but ... the URL parameters, are they really really necessary? Only asking because if you want to delete them later, Giggity uses the full URL as the primary key so doing so would result in 2 hackmas menu entries with separate selections and all in all terrible UX. I've started working out a fix but it's not there yet... :(

FWIW if I fetch the same URL without parameters I seem to get the exact same results at least, so I'm tempted to cut it off.

ruru4143:
The limit is necessary, because the default limit is something around 200-300. We won't change the parameters anymore tho.

ruru4143:
I really don't understand how these versionnumbers work...

ruru4143:
ok now i get it, it's current date + a number which must be growing

Wilm0r:
Yeah, sorry about that, I should remove version numbers from these src files too some day. Much of Giggity is 15yo, not all design decisions are great. :)

ruru4143:
Hahaha, sounds like it should get a rewrite at some point. Thanks for supporting this app for so long!

PR #371, opened 2024-08-19
Add Kieler Open Source und Linuxtage 2024

Moini:
Hi, @Wilm0r - it's that time of the year again :) 

Could you please add our event to the giggity list of events?

Have a great week!

### DIFF
--- a/menu/kielux_lpd_2024.json
+++ b/menu/kielux_lpd_2024.json
@@ -1,0 +1,28 @@
+{
+	"version": 2024082001,
+	"title": "Kielux Linux Presentation Day",
+	"url": "https://kielux.de/content/calendars/LPD24.ics",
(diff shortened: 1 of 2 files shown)


menu/kielux_lpd_2024.json
```diff
@@ -0,0 +1,28 @@
+{
+	"version": 2024081901,
+	"title": "Kielux Linux Presentation Day",
```
Wilm0r, line 3: Shouldn't there be a year in here BTW? Hm I see it was missing last 2 years as well but present before. While the menu does show the date in subscript I think it's still better to include a year here too..

Moini: It would have been okay either way, but now I added a year. Thanks for reviewing, @Wilm0r !

PR #372, opened 2024-08-22
State of the Map 2024

Zverik:
Same as always: a hourly updated merged schedule from https://sotm.osmz.ru/sotm2024.html

Also fixing a small deprecation in the ci script.


Wilm0r:
Cool, thank you :)

PR #380, opened 2024-09-14
 add Linux Plumbers Conference 2024 and a note about Indico event URLs

t-8ch:


README.md
```diff
@@ -77,6 +77,10 @@ It is usually in the form: `https://<CFPHOST>/<CONFERENCE>/schedule/export/sched
 
 It usually is in the form: `https://<CFPHOST>/conferences/<CONFERENCE>/schedule.xml`
 
+### Indico
+
+It usually is in the form: `https://<CFPHOST>/event/<CONFERENCE>/event.ics?detail=contributions`
```
Wilm0r, line 82: They only have .ics exports? :(

t-8ch: I think so, or their custom json format.
See also https://github.com/indico/indico/issues/6199

Wilm0r: Gotcha, guess I'll see how simple it is and maybe add support for it some day if it's worth the trouble. Just really hate `.ics` files :)

PR #384, opened 2024-10-02
Create bitcoin_amsterdam_2024

AbelLykens:
Bitcoin conference next week


AbelLykens:
Files placed correctly now, wonder how I trigger a re-run of the github action, maybe by posting here?

Wilm0r:
The mouseover on the https://github.com/Wilm0r/giggity/actions/runs/11140189517/job/30958420674?pr=384 timestamp says 04:12 for me which is 1min after your most recent comment, plus obviously yes it passed now, so we're good :)

AbelLykens:
Thanks! Yeah I noticed the comment did not trigger the new run, but changing the file did :).

Thanks for merging.

PR #385, opened 2024-10-03
added mrmcd2024

MoKe12g:
added mrmcd2024 event in Darmstadt, Germany

### DIFF
--- a/menu/mrmcd_2024.json
+++ b/menu/mrmcd_2024.json
@@ -17,11 +17,11 @@
 				"title": "Tickets"
 			},
 			{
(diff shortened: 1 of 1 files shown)


MoKe12g:
would use https://raw.githubusercontent.com/johnjohndoe/CampFahrplan/217bebab72951f1586704bca07ee8b3209f7b683/app/src/mrmcd2024/ic_launcher-playstore.png as logo, which should also have an alpha layer but seems too github dependent.

Wilm0r:
Sorry for not seeing this one in time! :( I guess I'll merge anyway so it could be used as a base for next year's entry..

PR #387, opened 2024-10-09
Add Ubuntu Summit 2024

z3ntu:
Plus a small fix in `tools/ggt.sh`

One thing to note, Android (13) doesn't seem to open ggt.gaa.st by default, it says something about the link not being verified so I assume this is something you need to do somewhere. However you can manually enable this in the app settings to still test this.


Wilm0r:
Cool yeah pretty sure that explains it. https://github.com/Wilm0r/giggity/issues/389 filed.

PR #390, opened 2024-10-10
Add LinuxDays 2024

oskar456:


Wilm0r:
Sorry for missing this :( Wondering whether I should figure out auto approvals. Frustrating that github doesn't have per-directory ACLs or so to make that easier. I guess let's merge for continuity anyway.

BTW in case you hadn't noticed, you *can* use `tools/ggt.sh` to generate a URL that includes all your JSON which should work for most people as a deeplink straight into Giggity, whether scanned as a QR or when linked to from the event website.

oskar456:
Sorry for late submission. Yes, we did have the deep link on our website, it worked great.

PR #391, opened 2024-10-18
Add IETF121

chrysn:
It's in 2 weeks time, so not sure if that makes sense still, or whether I'd better keep the file somewhere else and disseminate through QR codes.

### DIFF
--- a/menu/ietf121.json
+++ b/menu/ietf121.json
@@ -1,15 +1,14 @@
 {
-	"version": 2024101000,
+	"version": 2024101001,
 	"url": "https://datatracker.ietf.org/meeting/121/agenda.ics",
 	"title": "IETF 121 Dublin",
 	"start": "2024-11-02",
 	"end": "2024-11-08",
 	"timezone": "Europe/Dublin",
-	"refresh_interval": 1800,
 	"metadata": {
 		"links": [
 			{
-				"url": "https://www.ietf.org/how/meetings/121/",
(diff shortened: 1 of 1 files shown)


chrysn:
The link I've fixing, also the version number (not sure I got it right, just followed the bot's directions).

The Refresh is odd: I specifically did check that the server does send a 304 on ETag match. (I didn't check for date based verifciation, but always found ETag to be more widespread and reliable anyway). I'm removing this, but we do have last minute changes, so I'd prefer to put it back in.

Wilm0r:
Feel free to put the low refresh timeout back then, as long as the server is fast. :) In fact new versions of Giggity have a better HTTP fetcher that should send ETag headers, pretty sure I've released that a while ago already. So it's just the CI that is a little behind on facts.

2w is plenty notice when I'm (able to) pay(ing) attention to this queue BTW, np. Once submitted this bit will appear in ggt.gaa.st/menu.json ~instantly and Giggity loads that file about once a day or immediately if you swipe down to refresh in the main screen.

(And ahh, IETF in Dublin... I never had it that close back when I lived there and still had an RFC to publish.. :) )

chrysn:
Updated as you suggested -- the 2 weeks became 2 days because I missed this, but as you said, it's going out fast.

Wilm0r:
D'oh, yep forgot it too. Merging.

PR #396, opened 2024-12-05
add everything open 2025

Mause:
this is the new incarnation of Linux Conf AU


Wilm0r:
Ha, LCA renamed (but colour scheme still looks familiar I *think*?), TIL. :) Thank you!

PR #397, opened 2024-12-12
Add 38C3

schildbach:
I left out c3nav for now.

### DIFF
--- a/menu/38c3.json
+++ b/menu/38c3.json
@@ -1,0 +1,24 @@
+{
+	"title": "38C3",
+	"metadata": {
+		"links": [
+			{
+				"url": "https://events.ccc.de/congress/2024/infos/startpage.html",
+				"title": "Website"
+			},
+			{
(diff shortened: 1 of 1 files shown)


schildbach:
I also removed `refresh_interval` because:

> ERROR: Schedule 38C3 refresh_interval set below 1d while server never sends HTTP 304s

schildbach:
@Wilm0r ready for review.

Wilm0r:
Hrmm.... Hadn't noticed the refresh interval thing before. I should fix that check, Giggity should be better at HTTP caching now (including ETags, which are more likely to work IIRC).

CCC does probably make last minute schedule changes somewhat regularly?

mmuman:
Oh cool, won't have to do it.
Yes there are usually many updates to the schedule…

mmuman:
It's sad the [full schedule](https://events.ccc.de/congress/2024/hub/en/fahrplan?mode=calendar&kind=official&set=kall#now) doesn't pull from an xml…

PR #406, opened 2025-03-06
Create bibliocon25.json

zuphilip:
Same procedure as last/every year ;-)

### DIFF
--- a/menu/bibliocon25.json
+++ b/menu/bibliocon25.json
@@ -1,6 +1,6 @@
 {
 	"version": 2025030601,
-	"url": "https://cloud.bib.uni-mannheim.de/s/MRwnBF84RfdJw55/download/bibliocon25.xml",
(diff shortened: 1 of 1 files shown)


Wilm0r:
I get this when I try to merge: "JSON.parse: unexpected character at line 1 column 1 of the JSON data"

I guess I'll let Github be its global single point of failure for a while and try again later today. :|

zuphilip:
Oh, that is strange. I have no idea what the problem could be. I just copied the text from last year and created a new file directly in GitHub in the browser. If it is easier you can also [download the file](https://github.com/zuphilip/giggity/blob/patch-2/menu/bibliocon25.json) and add it directly to the master branch yourself.

Wilm0r:
Oh, no, it was an internal github error, the "JSON" wasn't referring to your PR contents at all I presume.. Anyway let's try, surely they've fixed this by now?

Wilm0r:
And thank you as last/every year! :)

PR #410, opened 2025-04-03
Add SSTIC 2025

mmuman:
Couldn't test here, Android 5 only and it says it's empty…


mmuman:
grrr redirections

Wilm0r:
Haha, sorry, do you think that check is too strict in today's day and age?

mmuman:
Dunno, in that case I think the redirection is probably temporary until the even is happening.

PR #413, opened 2025-04-25
Add JdLL 2025

mmuman:
Icon probably won't work, we only accept PNGs, right?


mmuman:
Of course not.

Wilm0r:
Well I guess theoretically a JPEG would work, but it seems ugly especially because it doesn't have an Alpha layer.

We could try and see what it looks like, but ... meh few people use the home shortcut feature anyway probably!

PR #414, opened 2025-04-28
FOSS4G Europe 2025

Zverik:
This adds an entry for FOSS4G Europe 2025 conference in Mostar. As usual, I've set up schedule merging.

Also I've updated `.gitignore` for some Python developer by-products, and added `pyproject.toml` if you don't mind, so one can do `uv run tools/menu-ci.py` and not deal with looking up dependencies.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 /app/build
 __pycache__
+.mypy_cache/
+.python-version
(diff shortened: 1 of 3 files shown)


Wilm0r:
Yep good idea re the `pyproject.toml`, thanks :)

PR #427, opened 2025-07-03
Add DebConf25

fortysixandtwo:
The logo is 404 currently, but should become available after https://salsa.debian.org/debconf-team/public/websites/dc25/-/merge_requests/50 is merged (and the homepage refreshed)

### DIFF
--- a/menu/debconf25.json
+++ b/menu/debconf25.json
@@ -1,0 +1,33 @@
+{
+    "version": 2025070302,
+    "url": "https://debconf25.debconf.org/schedule/pentabarf.xml",
+    "title": "DebConf 25",
+    "start": "2025-07-14",
+    "end": "2025-07-20",
+    "timezone": "Europe/Paris",
+    "metadata": {
+        "icon": "https://debconf25.debconf.org/static/img/logo-medium.d4e8d313ad22.png",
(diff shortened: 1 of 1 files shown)


fortysixandtwo:
Yes, the map would be super nice,
however, I think it starts complaining once there are more than 4 links, so we should remove one then.

How about putting "Map" in place of "DebCamp" ?

fortysixandtwo:
Right now validation still fails because of the image and adding another link would indeed also break validation.

```
Icon for DebConf 25 seems bad: 600×270 image is not square, 600×270 image is too large
5 links exceeds the recommended maximum of 4 to ensure the nav drawer fits on most screens with no scrolling
```

@stefanor About that image:
It's failing because the logo is
a) too large and
b) not square, do you happen to know if we have square logo's (PNG with alpha channel) on the homepage already?

stefanor:
How about https://debconf25.debconf.org/static/img/logo-medium.d4e8d313ad22.png

stefanor:
Actually, that doesn't look great, let me make it bigger

Wilm0r:
It's not a very  hard limit :) If you think they're all really useful let's keep them.

PR #428, opened 2025-07-03
Fix Debconf 25 indentation

NcamGnrvngu:
The broken indentation of Debconf 25 also affects the merger which updates the menu.json on ggt.gaa.st. This provides an easy fix so that the ci/cd can run again

### DIFF
--- a/menu/debconf25.json
+++ b/menu/debconf25.json
@@ -20,7 +20,7 @@
                 "url": "https://debconf25.debconf.org/about/coc/",
                 "title": "Code of Conduct"
             },
-	                {
+            {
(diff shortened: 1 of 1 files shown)


Wilm0r:
Whoops what have I done. Let me try to find the logs. ggt.gaa.st isn't meant to be this strict.

Wilm0r:
OK probably the CI will pass if you rebase, but in case you're not around, let me merge and I think ggt.gaa.st will come back. (Its GA/Py runtime is outdated I guess I'll need to do some more work to update that one...)

PR #429, opened 2025-07-15
Add HOPE 16 (2025)

mmuman:
Schedule is not yet online.


mmuman:
Schedule seems to have been published!

PR #433, opened 2025-07-30
Correct schedule URL for WHY2025 and add links.

eriktews:
The WHY2025 program was moved to https://program.why2025.org/


Wilm0r:
Oh balls. :( I hope that isn't going to ruin anybody's selections. I really need to fix that "URL is primary key" design flaw...

Wilm0r:
Merged. Oh ah yes I forgot, this is going to create a duplicate WHY2025 entry in the menu which is even more confusing. I guess let me retitle it. Sorry folks, if anybody sees it. I should fix that dumb bug. :|

PR #436, opened 2025-08-14
Kieler Open Source und Linux Tage 2025

Moini:
Hi, @Wilm0r !

It's the time of the year again - could you please add the Kielux schedule to Giggity?

### DIFF
--- a/menu/kielux_2025.json
+++ b/menu/kielux_2025.json
@@ -1,0 +1,48 @@
+{
+	"version": 2025081401,
+	"title": "23. Kieler Open Source und Linux Tage",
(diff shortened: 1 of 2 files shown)


Moini:
I would like to make a short instruction for Kielux visitors about how to get the schedule on their Android phone. 

Do you have an SVG (or at least higher resolution) version of the Giggity logo anywhere, @Wilm0r ? All I found in the repository was this tiny image: 

deoxide_icon

Wilm0r:
Oh yes, found it! https://github.com/Wilm0r/giggity/blob/master/gimp/deoxide_speaker_icon.svg

Moini:
Thanks :)

PR #441, opened 2025-08-26
Add End Summer Camp {21} (aka 2K25)

LucentW:
Always used Giggity when going at international events... now that we have our XML it's time for us to be present! :)

Thank you very much in advance.


Wilm0r:
Sorry for not catching this one in time!

Thanks for using Giggity, and yesss for using the XML format (.ics files are kind of sad) :)

PR #444, opened 2025-09-03
Create camerajapan2025rotterdam.json

sigio:
Don't know if a Filmfestival fits for giggity.
But i'd like to have it in there ;)

Schedule was converted by AI from the PDF, but looks to be correct. Will do manual adjustments if I find errors.

### DIFF
--- a/menu/camerajapan2025rotterdam.json
+++ b/menu/camerajapan2025rotterdam.json
@@ -12,8 +12,8 @@
 				"title": "Website"
 			},
 			{
-				"url": "PDF Schedule (source)",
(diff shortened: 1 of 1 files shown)


sigio:
will make more edits first...

Wilm0r:
I thiiiink somewhere in the source tree I do have a Python script that can emit the right format so then you just need to implement a parser that puts things in the right shape of Python dicts or classes. Maybe it helps.

Wilm0r:
Oh apparently I deleted it in 76c8dbfdf4600069f63ff4c7fa286e084daeaeba

Not sure why. There was a past non-Frab XML format that Giggity used in early versions, but that's not what that generator emits .... maybe I mixed things up.

PR #457, opened 2025-10-30
[2025-11-01] Correct title casing

jmlich:
The name OpenAlt is a combination of the words Open and Alternatives. It’s not an abbreviation, and the correct capitalization is OpenAlt, not OpenALT.

### DIFF
--- a/menu/openalt_2025.json
+++ b/menu/openalt_2025.json
@@ -1,5 +1,5 @@
 {
-	"version": 2025102700,
+	"version": 2025103000,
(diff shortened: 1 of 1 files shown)


Wilm0r:
Oh, good point, thank you :) In the past Giggity would automatically overwrite the menu title with what's in the schedule file but I remember now, I've removed that behaviour since sometimes the title in the event file is bad (just a URL, or VERY long, etc.).

BTW I've added a checker for this and it's not completely working, would you mind to rebase this PR? I hope to find out what keep it from working properly...

PR #459, opened 2025-11-27
[2025-12-08] Add Open Source Summit Japan 2025

z3ntu:


z3ntu:
The URL is accessible for me, not sure why it says HTTP 403?

Also fwiw it's already November 28th in Japan (and UTC actually as well) so the version check is failing a bit unexpected but I guess it'll be resolved on its own soon.

Not sure what to do about the "entry mismatch", "ossjapan2025" isn't a particularly nice title.

Wilm0r:
Yeah it runs in UTC time I think. UTC midnight has passed by now I think so let's rerun.

As for the title, yep that one's already marked as an info/warning message, indeed ignore it. :)

Wilm0r:
```
New: Open Source Summit Japan 2025
Fetching https://ossjapan2025.sched.com/all.ics
Fetching https://events.linuxfoundation.org/open-source-summit-japan/
Unchanged: FOSDEM 2026
File passed validation! \o/
```

The title message is absent because that one depends on an emulator connection which I didn't set up for my local run. Curious why the CI got a HTTP 403 twice when I didn't. Guess it's IP-address-based then not User-Agent. :-/

Anyway, this should all be fine, let's merge. And indeed the serial# error is gone.

PR #482, opened 2026-03-23
[2026-06-03] Add SSTIC 2026

mmuman:
Still can't test here on Android 5 ("appears empty").

### DIFF
--- a/menu/sstic_2026.json
+++ b/menu/sstic_2026.json
@@ -1,0 +1,16 @@
+{
+	"version": 2026032300,
+	"url": "https://www.sstic.org/2026/programme.ics",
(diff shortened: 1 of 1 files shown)


mmuman:
[I guess they didn't change their mind](https://infosec.exchange/@sstic/112534199552420089) and won't add the LOCATION anyway.

mmuman:
Maybe we could add a JSON field to force it?

Wilm0r:
Huh oh the date passed... Sorry

> Maybe we could add a JSON field to force it?

Hm, how would that work? I'm reluctant to add hacks just to deal with an already inferior file format. :(

PR #492, opened 2026-04-30
[2026-07-02] Add Camp Interhack 2026

mmuman:


Wilm0r:
Don't worry about the icon, the file is only 30k so who cares, the purpose of that check is to avoid downloading 1MiB icons so really I should've made it a file size check instead of being so rude to high-dpi screens. ;)

Date mismatch is odd though, is that expected?

mmuman:
Yeah officially it's 2-5th but I think there's just no event in the schedule yet for Sunday.

PR #495, opened 2026-05-08
[2026-05-15] Add TDF5'2026

chris-se:


Wilm0r:
That "entry mismatch" is informational, as you see CI did approve :) Though I think this title is so long it'll get broken in two which doesn't look too nice, how about removing Tübingen? Website doesn't seem to include that as part of the event name.

chris-se:
Sure, I've updated this, many thanks for the quick reply!

Wilm0r:
np, thank you for the contribution!

PR #496, opened 2026-05-11
[2026-05-27] Add Embedded Recipes 2026

z3ntu:


z3ntu:
> Error: Icon for Embedded Recipes 2026 seems bad: 1772×1772 image is too large

I'm not aware of a smaller hosted version. Should I just remove it?

> Entry mismatch: title='Embedded Recipes 2026' in entry ≠ 'ER2026' in schedule?

That's on purpose, the website very much talks about "Embedded Recipes 2026", not using the abbreviation there.

Wilm0r:
Yeah no problem. :) The icon check really should switch over to a file size check, the goal is to avoid large ~blocking downloads on slow conference WiFi but that shouldn't really be a problem here.

And the title check is especially optional yeah.

z3ntu:
Thank you!

PR #498, opened 2026-05-17
[2026-06-04] Add GPN24

NcamGnrvngu:

PR #500, opened 2026-05-30
[2026-06-04] fix: use the gpn24 schedule.xml with everything

margau:
This is a merged schedule.xml for gpn24, including the lounge schedule (and possibly other events like gulasch kitchen etc.)

I hope I got the Version field right?

### DIFF
--- a/menu/gpn24.json
+++ b/menu/gpn24.json
@@ -1,7 +1,7 @@
 {
-  "version": 2026051700,
-  "url": "https://cfp.gulas.ch/gpn24/schedule/export/schedule.xml",
-  "title": "24. Gulaschprogrammiernacht",
+  "version": 2026053000,
+  "url": "https://cfp.gulas.ch/merged/GPN24/everything.schedule.xml",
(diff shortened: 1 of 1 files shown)


Wilm0r:
> I hope I got the Version field right?

Yes :) And the errors above can be ignored, in https://github.com/Wilm0r/giggity/pull/498 we noticed the server has unfriendly user agent checks.

Old versions of Giggity didn't deal very well with URL changes like this, I've fixed that a few months ago and I hope most people have picked up that change. Those who did not will end up with two GPN schedules in their menu now. I wonder whether we should add a brief "(merged)" tag to the title to help those?

margau:
Thanks for the feedback, I've added the suffix - I hope it's clear now!

Wilm0r:
Yes, and since this is the first PR modifying the schedule URL like this I've done a little more testing and it does seem to work, including preservation of selected events (assuming IDs remain consistent of course). Terrific. :) Thank you!

PR #502, opened 2026-06-07
[2026-06-29] Add FOSS4G Europe 2026

Zverik:
Adding my repack for the [FOSS4G Europe 2026 schedule](https://2026.europe.foss4g.org/schedule/). Includes workshops, coffee breaks and side events. Basically the same I've been doing for years :)

Also updated dependencies in the `pyproject.toml`.


Zverik:
(the entry validates fine locally, it's another case of "fvsr: INVALID_ARGUMENT: Application credential header not valid. Please fix the client to pass a valid application credential header." in the CI)

Wilm0r:
What's up with the unrelated toml change though BTW?

Wilm0r:
And https://sotm.osmz.ru/fe2026.html is an actual HTML file as the extension already suggests.

Zverik:
Wow, good catch! I have fixed the extension, and added a check to `menu-ci.py` to catch those mismatching file types, given it downloads the file. I'm extracting other files to another pull request.

Wilm0r:
Yay green!

The error from the verbose CI logs was unlikely to be related BTW, the annoying problem with `adb logcat` is filtering by package name is a PITA, see https://stackoverflow.com/questions/6854127/filter-logcat-to-get-only-the-messages-from-my-application-in-android for an example of how to do it. Maybe I should try it anyway but what if the PID changes during an unexpected reload, etc :|

PR #504, opened 2026-06-09
Update pyproject.toml and add a check for non-XML/ICS files

Zverik:
subj. Would allow running `uv run tools/menu-ci.py` and also catch issues like I had in #502.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,13 @@
 [project]
 name = "giggity"
-version = "0.1.0"
+version = "0.2.0"
 description = "Giggity Tools"
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "dulwich>=0.22.8",
-    "jsonschema>=4.23.0",
-    "pillow>=11.2.1",
-    "urllib3>=2.4.0",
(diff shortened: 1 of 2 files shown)


Wilm0r:
Tricky thing about the dependencies is, a few months ago I upgraded `ggt.gaa.st` to a recent Dulwich version and it broke. I need to figure out that backward incompatibility issue before I upgrade.

Sucks given the security-related PR, but from what I can tell none of the issues affect the tiny fraction of Dulwich that I'm actually using. Still, I should try to clean that up soon...

Zverik:
Right, I've downgraded dunwich then. Looks like 0.24 should be compatible, and it fixes support for Python 3.9 according to its changelog.

Wilm0r:
Which of the dependency version bumps is load-bearaing for you then? You mean the existing versions break with more recent Python versions?

Zverik:
Forgot to answer your question — I had trouble because of the old `jsonschema` dependency, which failed to compile. Others I increased just because I checked that they had new versions. I guess the `toml` change doesn't do much for other scripts, so it won't affect the website or anything. Thanks for merging!

PR #507, opened 2026-06-16
[2026-06-25] Add openSUSE Conference 2026

aylen384:


Wilm0r:
Sorry for the CI issue, could you please rebase? I think it should work again now.

Wilm0r:
The 403 might be real, the other error is most likely false alarm, I'll check tomorrow.

aylen384:
I can open the link - maybe the network of the ci is blockd?

Wilm0r:
It's either the network or the User-Agent header. Ran into that recently already. Maybe a simple attempt at ditching all the AI scrape traffic.

aylen384:
As the 403 was only on the Website link, I removed it. (It seems that our AI crawler protection is blocking that, but not blocking the .XML path as it is whitelisted).

But there is still an error reported, because 53ae2e4ade49afa056a8b88c3d4ea360ddc9a829 checks for `<?xml` but our xml file begins with just the `<schedule>` tag.

aylen384:
The remaining error is bacause of a bug in menu-ci.py where it doesn't accept XML files, if they don't have the `<?xml` header.
Fix: https://github.com/Wilm0r/giggity/pull/509

Wilm0r:
Yeah I was worried that that check was too much. I've made a fix similar to yours. Also, in the end CI results don't block merges so let's go :)

Wilm0r:
Oh and feel free to send another PR for re-adding the website link. I wonder whether I could make CI more resilient, did you try running the CI locally and how did that go? (I'm wondering whether it's HTTP Header based or IP based)

aylen384:
I did run the CI locally (I'm pretty sure the way I ran in was not the best way to call it, though).
I ran into the issue locally as well.
Even when fetching the URL using `import requests; requests.get("https://events.opensuse.org/conferences/oSC26/")`.
Curl on the other side worked fine.
But it is not the user-agent that triggers the block. Python-requests with `User-Agent` set to `curl/1.2.3` is still blocked. IDK what exactly is checked, though.

PR #509, opened 2026-06-17
Fix tools/menu-ci.py: Accept XML files without <?xml tag

aylen384:


Wilm0r:
I've pushed this as 9e6b6b7 instead.

PR #512, opened 2026-06-25
[2026-07-03] Add Fantastischer Aalen Termin (FAT) 26

t-8ch:


t-8ch:
I can't reproduce that 404 :-/

Wilm0r:
> I can't reproduce that 404 :-/

Even weirder is that it seems to be intermittent? As-in no errors here now, yet now we're getting the error on your other PR...

Definitely something we can ignore but of course I hope Giggity itself won't run into these errors..

t-8ch:
It was not intermittent. This issue was that the c3nav validation was broken.
See #513. I dropped the c3nav bits from this PR, so it should be good to go.

PR #513, opened 2026-06-27
[2026-07-03] menu-ci: Use c3nav v2 API for validation

t-8ch:
The old "legacy API" was removed from c3nav.
Switch to the new v2 API.

### DIFF
--- a/tools/menu-ci.py
+++ b/tools/menu-ci.py
@@ -139,7 +139,7 @@ class HTTP():
 
 		self.hcache = {}
 
-	def fetch(self, url: str, img: bool=False, head: bool=False):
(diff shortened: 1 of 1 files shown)


Wilm0r:
Could you reduce this to just the CI change?

t-8ch:
Yes, the non-CI change was only as reproducer. Please ignore this PR until I remove the 'draft' state.

t-8ch:
I think this works now.

Wilm0r:
And I hope no Giggity-side changes are needed. Since those are just direct links to c3nav in the browser with no "real" API use (IIRC) I hope we're fine.

PR #517, opened 2026-07-03
IETF 127 Vienna (in two weeks!) and 127

arnt:


arnt:
Oh, I see, the checker fails because the schedule for 127 is empty still.

PR #518, opened 2026-07-03
[2026-07-18] IETF 126, starting in two weeks

arnt:
I'll try to add 127 as soon as the agenda is published.

(I'm a little late this time.)


arnt:
That was quick! Thanks so much!

PR #524, opened 2026-07-21
[2026-11-14] Add Capitole du libre 2026

mmuman:


mmuman:
I get "Failed SSL handshake" on Android 5, I guess the root certificate workaround doesn't work anymore, or I need to add another one… or it's because of TLS 1.3 :-(

mmuman:
Looks like CI went nuts?

Wilm0r:
Hah that's fun, it ran on two PRs almost simultaneously and yes the bit that asks a live Android emulator to load it has no locking scheme. Yes feel free to mock me for that design flaw but this is a hobby project. :) Let's re-trigger.

Wilm0r:
And yes that serial number stuff needs to go away already...

PR #533, opened 2026-08-07
Fix events crossing the day-change boundary disappearing entirely

### LINKED ISSUE #532: Room (and associated session) not displayed
We created PR #531 that was merged into main to add RomHack Camp 2026 event.

While checking the app on my android phone I see RomHack Camp 2026 listed but I see only one room (STAGE 1) listed

Image

while the XML schedule file https://cfp.romhack.io/romhack-camp-2026/schedule/export/schedule.xml contains multiple rooms (2 at the moment) and one sessions associated with the 2nd room (called WORKSHOP2)

Image

I expect to see also WORKSHOP 2 and the associated session in the app

CyberSaiyanIT:
Fixes #532

## Bug

`Schedule.java` assigns each event to a "day" (`[dayChange, dayChange+24h)`, `dayChange` defaults to 06:00) by requiring **both** the event's start and end time to fit inside that window. An event that starts on one day but runs past the *next* day's boundary matches neither adjacent day, so it — and any room that only contains such an event — silently disappears from every view.

This is exactly what happened with RomHack Camp 2026: the "Community CTF by fibonhack" event in the `WORKSHOP 2` room runs from `2026-10-02T17:00` to `2026-10-03T08:00` (15h), which crosses the default 06:00 day-change. Its end time falls after day 1's window closes, and its start time falls before day 2's window opens, so it's dropped from both — taking the entire `WORKSHOP 2` room down with it, since that event was its only entry that day.

## Fix

Assign each event to the day its **start** time falls in, regardless of where its end time lands. This matches how `BlockSchedule`/`BlockScheduleVertical` already render items using their absolute start/end time rather than clipping to the day window, so the event still renders fully (just extending that day's grid a bit further), it just now gets bucketed into a day at all.

Changed in three spots in `Schedule.java` that all used the same start+end containment check:
- day-list construction (which days exist at all)
- `setDay()`'s `dayFirstTime`/`dayLastTime` computation
- `ItemList.getItems()`'s per-day filtering

## Testing

Added `testRomhackCamp2026` to `ScheduleTest.java`, using the actual RomHack Camp 2026 `schedule.xml` as a fixture (`app/src/test/resources/romhack_camp_2026.xml`). It asserts `WORKSHOP 2` and its CTF event show up on day 1.

- Verified the new test fails on the old code (reproducing #532) and passes with the fix.
- Ran the full `ScheduleTest` suite before and after: no new failures. (`test36c3`'s language-count assertions fail identically on unmodified `master` too — pre-existing, unrelated to this change.)


Wilm0r:
Let me dig through the fix later. But keep in mind that existing Giggity versions will still fail to load your schedule ... would you be able to set the day change time in the RomHack schedule to 08:00 to see whether that also helps for people who won't have upgraded yet?

CyberSaiyanIT:
> Let me dig through the fix later. But keep in mind that existing Giggity versions will still fail to load your schedule ... would you be able to set the day change time in the RomHack schedule to 08:00 to see whether that also helps for people who won't have upgraded yet?

Let first validate the PR and see if it works. I'm not ok with cahnging the schedule because the CFP is a single event that spans over 2 days and we want to keep it as-is
But as I said let's first check if the fix works :)

Wilm0r:
Sorry for the delay, was doing only simple menu PRs for a while as I'm going through an intercontinental move + job change and some more things all at the same time this Summer. :)

I'm still not a huge fan of this change when the real fix is simply to set the day change correctly in the upstream schedule. Right now the schedule is simply invalid.

OTOH the change is mostly quite innocent in the spirit of "liberal in what you accept, strict in what you send". There could be events that go on day and night and therefore even need this behaviour to change? And from memory your implementation doesn't end up including an affected event on multiple days which is indeed how I think it should be.

But what will the block schedule look like if someone now sneaks a year-long event in it? How long to accept? Probably still 24 hours?

I'm curious to see what other parsers do with it but I suppose at least some of them indeed do accept this already, too.

CyberSaiyanIT:
Ok, i understand the rationale behind the choice. I asked also Hacker tracker to add RomHack Camp schedule to the app, I'll let you know how HT behaves on events that span multiple days

CyberSaiyanIT:
We managed to have RomHack Camp included in Hacker Tracker and as you can see the CTF is visible
aa7994cc-55ff-483a-a661-360c4a22bd41
c6ec0710-71ac-4674-8375-796c0534d946

CyberSaiyanIT:
I just noticed that another event (Friday music session, same for Saturday music session) that starts on friday and ends on saturday at 00:30 is visible. I expect the same behavior of the CTF event for these events

2e5270e1-8773-4a0f-aeb6-b85e4fb14f18
b8fde927-2b94-4873-8b1e-a6404741b3bf

PR #536, opened 2026-08-15
[2026-08-28] Add State of the Map 2026

Zverik:
Same as before, adding our mapping conference.

### DIFF
--- a/menu/sotm_2026.json
+++ b/menu/sotm_2026.json
@@ -1,0 +1,17 @@
+{
+    "version": 2026081600,
+    "url": "https://sotm.osmz.ru/sotm2026.xml",
+    "title": "State of the Map 2026",
+    "start": "2026-08-28",
+    "end": "2026-08-30",
+    "timezone": "Europe/Paris",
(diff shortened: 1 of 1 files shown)


Wilm0r:
A schedule on pretalx.com does always look more authentic but yes I know you're usually the SotM submitter :) How did this fragmentation of schedules happen BTW? And curious, are IDs consistent between the two?

Zverik:
Hey Wilmer, the reason is pretty simple: SotM is a big conference consisting of multiple parts (just like FOSS4G I'm also managing), so it has more than one Pretalx source: namely, OSM Science rooms. I have written a script to merge multiple source together, and that's why we have a separate website to host the result. You can see the script downloading the sources in [my Ansible playbook](https://github.com/Zverik/ansible-textual/blob/main/roles/schedules/files/update_schedules.sh).

I did this PR specifically after a member of SotM program team reminded me, so consider this one official :)

Wilm0r:
Heh, understood. Thank you for the background!

PR #537, opened 2026-08-17
[2026-09-01] End Summer Camp {22} / 2k26

LucentW:
And here's this year event too :wink:

### DIFF
--- a/menu/esc_2k26.json
+++ b/menu/esc_2k26.json
@@ -1,0 +1,28 @@
+{
+	"version": 2026081700,
+	"url": "https://pretalx.endsummer.camp/2k26/schedule/export/schedule.xml",
+	"title": "End Summer Camp {22}",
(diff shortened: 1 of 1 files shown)


Wilm0r:
You can ignore the title mismatch and the link redirect, that last check I should probably just delete. But is the start date discrepancy expected?

LucentW:
Hi @Wilm0r,
the event officially starts on September 1st, I'll be there to setup A/V and stuff and a handful of other folks doing general event setup: we're small enough that buildup is 100% community based, but we (currently) have no talks/lab on schedule for day 1, so the XML has the `<day>` node, albeit with no `<room>` and `<event>` children :)

Thank you!

Wilm0r:
So see https://github.com/Wilm0r/giggity/pull/539 near the end, Giggity might still overwrite these. But I guess it's not a big deal.

PR #538, opened 2026-08-19
[2026-08-20] [2026-08-26] [2026-08-27] Create hackmas26.json

1u:
Thx for the existence of Giggity

### DIFF
--- a/menu/hackmas26.json
+++ b/menu/hackmas26.json
@@ -14,8 +14,8 @@
     },
     "url": "https://openki.hack-mas.at/api/0/xml/schedule?region=6JRpcbs8v39PyeA48&limit=999",
     "timezone": "Europe/Vienna",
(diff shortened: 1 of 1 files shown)


1u:
Validation check was cancelled after 6h. That all right or do I need to change something?

Wilm0r:
Yeah the thing's a little fragile, sorry. :(

Can you fix the dates and the timezone?

1u:
Yaay :)

Wilm0r:
\o/

And thank you for contributing!

Also, I see this is another interesting case of the event start date not really matching the schedule file, I wonder whether I should relax that check too..

1u:
I don't have strong opinion on this. Maybe have two states in the check like "fails" and "warnings"?

liebe Gruess us Züri

PR #539, opened 2026-08-20
[2026-08-30] added FOSS4G 2026 schedule

lucadelu:


lucadelu:
@Wilm0r the dates mismatch is correct, I think now the file is ready to be approved

Wilm0r:
In the past Giggity would still overwrite `menu.json` start/end dates with dates from the schedule after the first load, not sure whether that's still the case by the way.

But either way, if you think these dates are more correct I'm not going to try to "correct" them. :)

PR #540, opened 2026-08-28
[2026-09-25] Add postmarketOS and Alpine Linux Conference 2026

z3ntu:


Wilm0r:
I'd love to take SVG files as icons but the Android API doesn't support this other than including them in src and converting to PNG at build time :( Have you got a PNG?

z3ntu:
Sidenote, I'm part of the organizers of this conference.

I've removed the icon for now, we actually don't really have a proper conference logo yet, so just using the postmarketOS logo (or the Alpine logo) wouldn't do it justice. If we get one later, I can add it then!

PR #543, opened 2026-09-07
Fix hashCode

jpantonow:
## What does this implement/fix?

| Property | Value |
|---|---|
| Method | `hashCode` |
| Class | `net.gaast.giggity.Schedule$Item` |
| File | `app/src/main/java/net/gaast/giggity/Schedule.java` |
| Specification | `MessageDigestSpec` |
| Analysis tool | `ape` |

### Observed dynamic-analysis failure

The dynamic-analysis run by `ape` reported (`MessageDigestSpec`) while exercising `net.gaast.giggity.Schedule$Item.hashCode`:

> expecting one of {SHA-256, SHA-384, SHA-512} but found MD5.

### Root cause

Schedule.Item.hashCode() invokes java.security.MessageDigest.getInstance("MD5"), and MD5 is outside the verified MessageDigestSpec allow-list.

### Correction strategy

Apply the authorized hashCode-override migration: eliminate MessageDigest rather than substituting another governed digest, and use an idiomatic non-cryptographic int hash over the same live data currently represented by getUrl() (Schedule URL and guid-or-id). Preserve the override signature and int result while treating any changed exact integer values as a compatibility item requiring the specified consumer and persistence review.

### Coordinated changes

- `app/src/main/java/net/gaast/giggity/Schedule.java`: ByteBuffer was used only to fold the first four bytes of the retired MD5 digest into the hashCode result.; MessageDigest and NoSuchAlgorithmException were used only by the retired MD5-based hashCode implementation.; Use the idiomatic non-cryptographic hash for the same sole input as before: getUrl(), which combines the enclosing schedule URL and guid-or-id discriminator. This preserves the public override signature and int result while eliminating MessageDigest, its algorithm fallback, and digest-byte folding.
- `app/src/test/java/net/gaast/giggity/ScheduleHashCodeTest.java`: Adds non-reflective JUnit coverage for the supplied cross-file contract: Schedule.Item instances with equal enclosing Schedule URLs and equal guid-or-id discriminator data must have equal hash codes, and equality must imply hash-code equality. The assertions deliberately do not depend on the prior MD5-derived integer.

### Verification included

- Run the relevant JVM unit-test target containing app/src/test/java/net/gaast/giggity/ScheduleHashCodeTest.java.
- Run the JVM unit-test target that includes app/src/test/java/net/gaast/giggity/ScheduleHashCodeTest.java.
- Verify items with equal Schedule URLs and equal guid-or-id discriminator values return equal hashCode values.
- Verify Schedule.Item equality implies equal hashCode values.

### Residual review requirements

The patch remains review-only until these assumptions are confirmed:

- The reviewed scope classifies this value as an internal runtime discriminator; no persisted, transmitted, exported, cross-release, or external exact-value consumer was identified.
- The existing super.hashCode() fallback demonstrates that callers already tolerate non-digest-derived and process-unstable values.
- getUrl() remains non-null for constructed Schedule.Item instances, as required by both the previous digest implementation and the replacement String.hashCode() call.
- The supplied target source is empty because this is a new file; production source was not included for direct API verification.
- This draft assumes the existing testable production API is new Schedule(String scheduleUrl).new Item(String guidOrId), as indicated by the target plan's reference to an existing Schedule/Item setup. Confirm the exact constructor visibility and discriminator semantics during review.
- The coordinated production edit preserves Schedule.Item.equals(Object) behavior based on the same getUrl()-participating fields, so equal instances constructed with equal schedule URLs and discriminator strings compare equal.

## How was this tested?

- [x] Automated test coverage added in 1 test file(s)
- [x] Project compiles successfully (`:app:testDebugUnitTest`)
- [ ] Confirmed by a project maintainer

## Reviewer notes

Do not merge until these assumptions and blockers are resolved:

- The reviewed scope classifies this value as an internal runtime discriminator; no persisted, transmitted, exported, cross-release, or external exact-value consumer was identified.
- The existing super.hashCode() fallback demonstrates that callers already tolerate non-digest-derived and process-unstable values.
- getUrl() remains non-null for constructed Schedule.Item instances, as required by both the previous digest implementation and the replacement String.hashCode() call.
- The supplied target source is empty because this is a new file; production source was not included for direct API verification.
- This draft assumes the existing testable production API is new Schedule(String scheduleUrl).new Item(String guidOrId), as indicated by the target plan's reference to an existing Schedule/Item setup. Confirm the exact constructor visibility and discriminator semantics during review.
- The coordinated production edit preserves Schedule.Item.equals(Object) behavior based on the same getUrl()-participating fields, so equal instances constructed with equal schedule URLs and discriminator strings compare equal.

## Additional context

I’m an undergraduate Computer Engineering student at the University of Brasília (UnB), and this contribution is part of a research project involving software security analysis.

I’m happy to adjust the implementation to better match the project’s architecture, coding conventions, or maintainers’ recommendations.


Wilm0r:
Oh, just made this change myself independently.

The hash used really doesn't matter but it's a nice cleanup anyway. Wish I remember why I even implemented that original contraption.

PR #544, opened 2026-09-11
[2026-10-21] add The Matrix Conference 2026

HarHarLinks:


HarHarLinks:
huh. ggt.sh works for me.

Wilm0r:
Yeah sorry, the verification against a live Giggity instance is still quite brittle..

HarHarLinks:
hm, technically the schedule contains "break" entries on the days the CI calls out. i'd love for these to also be displayed in the app, but i think this isn't actually supported? if that's correct i can adjust the files.

Wilm0r:
Are those breakouts not in the Frab XML file?  Or is there a parsing issue like https://github.com/Wilm0r/giggity/pull/533 maybe?

HarHarLinks:
not breakouts, but what pretalx calls "breaks", i.e. labels for pauses that you can also use "creatively". frab xml appears to not support this and/or pretalx doesn't include them. see https://cfp.2026.matrix.org/matrix-conference-2026/schedule/

Wilm0r:
Funny, the 20th really is in the XML file even but as an empty day: `<day index="1" date="2026-10-20" start="2026-10-20T04:00:00+02:00" end="2026-10-21T03:59:00+02:00"> </day>`

which indeed gets ignored by Giggity since the dates are pulled from actual event timestamps not from any raw XML data.

Anyway, if you say the 20th is really the start date you want to show then feel free to keep that. I'll give you a moment to update the PR. The CI doesn't get to block merges, it's advisory only. :)

HarHarLinks:
Thanks, I reckon it's at least good enough :)

PR #545, opened 2026-09-15
[2026-10-05] [2026-10-07] [2025-10-07] Add Open Source Summit Europe 2026

okias:


Wilm0r:
Hmm interesting, another case of the CI disagreeing with you on the event start/end date? Indeed the webpage says it starts on the 7th while the .ics has events on the 5th already, do you know why? Just like in https://github.com/Wilm0r/giggity/pull/544 , it's OK to ignore that CI warning by the way if you think you're right.

okias:
yeah, there are associated events starting earlier, but the main conf really starts 7th, I'll update

Wilm0r:
Just verified: Yes after the first load of the schedule file it overwrites these dates with those from the schedule, so then it'll still show up as 5–9 Oct.

Don't think I'll change that behaviour, it's a bit confusing but really I don't know what's truly correct in cases like this.

